PR #2, opened 2015-04-13
Added a travis.yml configuration.

squidarth:
Added a travis.yml configuration.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,24 @@
+language: python
+python:
+    - 3.4
+    - 3.3
+    - 2.7
+sudo: false
+before_install:
+    - git clone --quiet --depth 1 https://github.com/minrk/travis-wheels travis-wheels
+    -  if [ ${TRAVIS_PYTHON_VERSION:0:1} == "2" ]; then wget http://repo.continuum.io/miniconda/Miniconda-latest-Linux-x86_64.sh -O miniconda.sh; else wget http://repo.continuum.io/miniconda/Miniconda3-latest-Linux-x86_64.sh -O miniconda.sh; fi
+    - chmod +x miniconda.sh
+    - ./miniconda.sh -b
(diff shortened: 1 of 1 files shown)


minrk:
You will need a requirements.txt to get the ipython/jupyter dependencies. See the one in jupyter_nbconvert for an example.

I'm not sure if this is going to work because apt-get install pyqt doesn't make pyqt available to the virtualenv where I think the tests are run.

takluyver:
One way we could do this is to install Miniconda and use that to install pyqt. It's kind of a hack, but it works. Here's an example that does it: https://github.com/jhamrick/dbtools/blob/master/.travis.yml

squidarth:
What did you do to install pyqt before this repo was split out?  Still a little stuck on the best approach and was digging through old versions of github.com/ipython/ipython

takluyver:
We didn't ;-) - the Qt console wasn't tested on Travis. We also run tests on our [Jenkins server](https://jenkins.jupyter.org/), where we can install Qt. But now that the Qt console is a separate repo, doing this doesn't hold up tests for the rest of IPython, so it makes sense.

squidarth:
@takluyver Finally got the tests to pass! Lmk what you think

Carreau:
Does it uses docker on travis ? Or should we set `sudo: false` ?

Also sidenote, I guess on docker build we could use cached dirs to speed-up rebuild.

squidarth:
Ah good catch--I was previously installing pyqt via `apt-get`. `sudo` can be set to `false`.  It's not using docker, although that would be a good thing to do as a follow-up

Carreau:
depends also on when the repo was created, I think they might default to sudo:false now.

Carreau:
decreased the time by ~30 sec /run roughly 1min15 total.

minrk:
Great, thanks for getting this started. Since it's using conda, you can probably remove the use of travis-wheels, and install a few dependencies with `conda install pyzmq sphinx pygments`.

squidarth:
Cool--went ahead and made that change

takluyver:
:+1:

minrk:
Great!

PR #4, opened 2015-04-19
deprecate IPythonQtConsole

### LINKED ISSUE #3: Apparently redundant parent App class
@minrk, in 129c5af467c46fe29cd6c489278f719d83683d55, you made the application class look like this:

```python
class JupyterQtConsoleApp(object):
    pass

class JupyterQtConsoleApp(JupyterApp, JupyterConsoleApp, JupyterQtConsoleApp):
    ...
```

Is there a good reason for this pattern, or is it a leftover from some debugging? If there's a reason, let's add a comment explaining it.

minrk:
Warn about use of deprecated IPythonQtConsole, keeping old config working.

closes #3

### DIFF
--- a/jupyter_qtconsole/console/qtconsoleapp.py
+++ b/jupyter_qtconsole/console/qtconsoleapp.py
@@ -43,6 +43,7 @@ if os.name == 'nt':
 
 from jupyter_qtconsole.qt import QtCore, QtGui
 
+from traitlets.config import Configurable
 from traitlets.config.application import boolean_flag
 from traitlets.config.application import catch_config_error
 from jupyter_qtconsole.console.ipython_widget import IPythonWidget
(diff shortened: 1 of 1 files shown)


takluyver:
This will be very confusing for any code that actually tries to import and use IPythonQtConsoleApp, because it will exist but won't do anything useful. Perhaps that's inevitable if we want to preserve config.

Given that the config file will need to be in a different place with a different name anyway, maybe we should just change the class name. We could provide a migration tool that moves the config files and does a find-replace IPythonQtConsoleApp -> JupyterQtConsoleApp.

minrk:
I realize I can do the deprecation in another, better way, anyway.

takluyver:
OK - I'll leave it up to you whether to merge this in the meantime or close it - it looks fine to me.

minrk:
IPythonQtConsole should be properly deprecated, with warnings, etc.

PR #5, opened 2015-05-07
s/jupyter_qtconsole/qtconsole/

minrk:
package is now just qtconsole


damianavila:
LGTM, and after installing some missing dependencies, I tested successfully...

PR #9, opened 2015-05-08
ipython_kernel is now ipykernel

minrk:
ipython/ipykernel#14


Carreau:
you missed the travis.yalm fiel which have ref to ipython_kernel.

PR #10, opened 2015-05-14
Flatten qtconsole

minrk:
remove qtconsole.console subpackage

The package only has one entrypoint.


qtconsole/tests/test_console_widget.py
```diff
@@ -5,7 +5,7 @@
 from qtconsole.qt import QtCore, QtGui
 
 # Local imports
-from qtconsole.console.console_widget import ConsoleWidget
```
takluyver: You took out a bit too much here.

takluyver:
This seems fine to me, modulo the typo that's causing a test failure.

damianavila:
LGTM too after fixing the `No module` error...

minrk:
Fixed.

Issue #11, opened 2015-05-27
The value of WM_CLASS in IPython qtconsole is empty.


183amir:
'WM_CLASS' variable is used in some desktop environments to associate apps with their corresponding '.desktop' file. IPython qtconsole's WM_CLASS seems to be empty.

    xprop WM_CLASS
    WM_CLASS(STRING) = "", ""

This makes the app to have no title in Gnome Shell.

Reference: https://wiki.gnome.org/Projects/GnomeShell/ApplicationBased

takluyver:
That page has an example to set it for PyGObject code. If there's a simple equivalent you can find for PyQt, pull requests are welcome.

183amir:
Please take a look at [here](http://www.qtcentre.org/threads/31012-How-to-change-default-windows-class-name), [here](http://stackoverflow.com/questions/11846029/customize-wm-name-and-wm-class-as-shown-by-xprop-in-pyqt4), and [here](https://bugreports.qt.io/browse/QTBUG-29013).
These are not solutions but they may be useful.

takluyver:
On my GNOME 3 desktop (Debian testing), I do already see a title:

![screenshot from 2015-05-27 11 03 37](https://cloud.githubusercontent.com/assets/327925/7843702/2af3006e-0460-11e5-8ce8-c9816051fe20.png)

ccordoba12:
@takluyver, I think the problem is with KDE.

takluyver:
The OP said "This makes the app to have no title in Gnome Shell."

ccordoba12:
Oops, sorry, didn't read that part. I said it because we had a report from a KDE guy about this problem a couple of months ago.

takluyver:
It's entirely possible that there's something similar going on in KDE.

damianavila:
Just to add... I am in gnome-shell since two days ago (still looking "my" perfect desk env), and I can see the logo as @takluyver mentioned...

damianavila:
But I can not see the title... at least with IPython 3.1.0:

![captura de pantalla de 2015-05-27 17 41 07](https://cloud.githubusercontent.com/assets/1640669/7847028/99f94b08-0497-11e5-8c25-73cf45575031.png)

takluyver:
@damianavila - are you launching that from a terminal, or from the launcher? What version of GNOME do you have? I seem to be on 3.14.4.

damianavila:
[...]

PR #12, opened 2015-05-27
Update qtconsoleapp.py

183amir:
Fixes jupyter/qtconsole#11
I did a quick test on my machine and it seems to work.

### DIFF
--- a/qtconsole/qtconsoleapp.py
+++ b/qtconsole/qtconsoleapp.py
@@ -218,7 +218,7 @@ class JupyterQtConsoleApp(JupyterApp, JupyterConsoleApp):
 
     def init_qt_app(self):
         # separate from qt_elements, because it must run first
-        self.app = QtGui.QApplication([])
+        self.app = QtGui.QApplication(['IPython'])
 
     def init_qt_elements(self):
         # Create the widget.


Carreau:
Hum, with the renaming to Jupyter, I'm not sure `IPython` a the right choice :-)

takluyver:
I think we need to work out what's going on a bit better:

- On my machine, I already see an application title in GNOME shell, without this change. Though it differs depending on if I start it from the command line or from a launcher. Which way are you starting it?
- The guide you posted said that the name needs to match the name of a `.desktop` file. That will probably be `ipython-qtconsole` or `ipython3-qtconsole`.
- @ccordoba12 reckons there's a similar issue on KDE; it would be good to know if that's the same or something else.

ccordoba12:
This is the issue reported in our side about our lack of WM_CLASS: spyder-ide/spyder#2207.

According to the OP (a KDE developer), it was causing Spyder to be grouped with the wrong app in Plasma's taskbar. So there are other side effects besides the one reported on issue #11.

takluyver:
OK, so it looks like we should be calling `self.app.setApplicationName()` as well. The [GNOME docs](https://wiki.gnome.org/Projects/GnomeShell/ApplicationBased) also say that the name should match the filename of the .desktop file, minus `.desktop`, which will be `jupyter-qtconsole` by the next release, I guess. I'm going to whistle up another PR with the changes I think are necessary.

takluyver:
See #16

PR #14, opened 2015-05-31
Document config options for Qt console

takluyver:
Using the newly merged traitlets machinery for this.

@jdfreder when we finalise this, it's probably worth doing something similar for the notebook docs. It looks like in jupyter/notebook#124, you've generated it and committed the output; this approach automatically generates it when building the docs, which should make things easier when we add or change options.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,7 +4,7 @@ dist
 _build
 docs/man/*.gz
 docs/source/api/generated
-docs/source/config/options
+docs/source/config_options.rst
 docs/source/interactive/magics-generated.txt
 docs/gh-pages
 jupyter_notebook/notebook/static/mathjax
--- a/docs/Makefile
+++ b/docs/Makefile
@@ -50,11 +50,16 @@ help:
 
 clean:
 	rm -rf $(BUILDDIR)/*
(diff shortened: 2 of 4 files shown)


jdfreder:
Sounds good to me.  Looking at this, it's pretty simple.  I'm fine with it as-is.  Was there something else you wanted to do with it?

takluyver:
No, I think it's fine :-)

jdfreder:
Great! Thanks!

takluyver:
Hmm, I've just realised that readthedocs doesn't run the Makefile but invokes Sphinx directly, so this won't work as is with docs on RTD (the equivalent for the IPython docs on RTD doesn't work either). That's annoying. I expect we can come up with some workaround, though.

jdfreder:
For now, we can just check the file in.  Do you want to open a PR doing that, and removing it from gitignore?

takluyver:
I suspect we can come up with a better workaround, like getting Sphinx's `conf.py` file to run the autogen script.

jdfreder:
https://github.com/jupyter/notebook/pull/124 implements the same mechanism you have here.  When a workaround is settled on, it will have to be implemented there too.

PR #15, opened 2015-06-01
Workaround to generate config docs on RTD

takluyver:
Including mocking out PyQt4 so it can import the Qt console files.

Ping @jdfreder

### DIFF
--- a/docs/autogen_config.py
+++ b/docs/autogen_config.py
@@ -1,5 +1,6 @@
 #!/usr/bin/env python
 
+import os.path
 from qtconsole.qtconsoleapp import JupyterQtConsoleApp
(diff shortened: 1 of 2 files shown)


jdfreder:
I don't understand what the mock stuff is for, and why it happens **after** the doc generating PY file is called.

takluyver:
It happens after because I'm being thick. One moment while I revise history...

It's there because we can't easily install pyqt on RTD, but without it we can't import the Qt console code to discover its config options. Mocking it out should make it possible to load the code - we don't need to run it, just load it.

takluyver:
Merging this, let's see if it works

PR #16, opened 2015-06-05
Set application name for Linux desktop frameworks

### LINKED ISSUE #11: The value of WM_CLASS in IPython qtconsole is empty.

takluyver:
From the information I've seen, I think this is right, but I can't test it on GNOME 3 for a few more days.

Closes #12
Closes #11

### DIFF
--- a/examples/jupyter-qtconsole.desktop
+++ b/examples/jupyter-qtconsole.desktop
@@ -3,22 +3,21 @@
 
 [Desktop Entry]
 Comment=Enhanced interactive Python qtconsole
-Exec=ipython qtconsole
+Exec=jupyter qtconsole
 GenericName[en_US]=Python shell
 GenericName=Python shell
 Icon=gnome-netstatus-idle
-Name[en_US]=IPython Qt console
-Name=IPython Qt console
+Name[en_US]=Jupyter Qt console
+Name=Jupyter Qt console
 Categories=Development;Utility;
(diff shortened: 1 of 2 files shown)


183amir:
Your `WM_CLASS` variable doesn't have to be lower case letters while your `.desktop` file name should be all lowercase. Usually, the `WM_CLASS` variable has the same case as application's name.

This code would work too (and it is usually done this way):

    self.app = QtGui.QApplication(['Jupyter-qtconsole'])
    self.app.setApplicationName('Jupyter-qtconsole')

takluyver:
But what's passed to `QApplication()` is usually the command line arguments, right? And commands are usually lower case.

183amir:
True. but how about `setApplicationName`?

takluyver:
Should they be different? Does the capital letter really matter anyway? Can you test this branch (with the `jupyter-qtconsole.desktop` file installed)?

takluyver:
This appears to be doing the right thing for me, so I'm merging it.

Issue #13, opened 2015-05-27
Docs


takluyver:
Making an issue so it doesn't get forgotten. Assign yourself if you want to work on this

There's a simple skeleton of docs there, but we should:

- [ ] Update stuff (e.g. it's now `jupyter qtconsole`)
- [ ] Create API docs (particularly considering the use case of embedding it as a Qt widget)
- [ ] Get it building on RTD

minrk:
We don't have API docs, yet. Marking that for 4.1

Carreau:
rtd now support conda and potentially rtd.yml cf https://github.com/rtfd/readthedocs.org/issues/857 might be woth looking into.

Carreau:
Docs page for rtd: 
  - http://docs.readthedocs.org/en/latest/conda.html
  - http://docs.readthedocs.org/en/latest/yaml-config.html

ccordoba12:
I don't think this is ready for 4.2. @takluyver?

takluyver:
This is not. If we're aiming to get 4.2 out soon, it's fine to bump this.

ccordoba12:
Yes, we are, specially for the fixes to Latex support.

takluyver:
OK. I've created a new 4.3 milestone and set it to that, on the grounds that it does still make sense to add API docs for a minor release if we do another 4.x release.

ccordoba12:
Ok. I just moved this to 5.0 because I didn't find another milestone available.

PR #17, opened 2015-06-08
Host built docs on Github pages

takluyver:
My efforts to mock out Qt on Readthedocs didn't work very well, so I think it's simplest to host these docs on gh-pages instead.

This will be at https://jupyter.github.io/qtconsole/dev/

I copied the gh-pages.py script from IPython, updated the repo and changed it to Python 3. I've commented out the command that tries to get the last 3 commits, because there aren't 3 commits yet, so that fails.

### DIFF
--- a/docs/Makefile
+++ b/docs/Makefile
@@ -24,6 +24,7 @@ I18NSPHINXOPTS  = $(PAPEROPT_$(PAPER)) $(SPHINXOPTS) source
 help:
 	@echo "Please use \`make <target>' where <target> is one of"
 	@echo "  html       to make standalone HTML files"
+	@echo "  gh-pages   clone qtconsole docs in ./gh-pages/ , build doc, autocommit"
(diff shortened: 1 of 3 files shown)


takluyver:
It looks like styling and images are missing, hang on

Carreau:
+1, want to set the theme as read the docs, and have travis build and push ?

minrk:
Sad, but understandable.

takluyver:
Styles and images are now [working](http://jupyter.org/qtconsole/dev/). The theme is currently Sphinx's new default 'Alabaster', which is quite nice, but I'm happy to change this to the RTD theme if we want that consistency. Let's leave automated building for another day.

Carreau:
Alabaster id fine :-)

Carreau:
there is a reference to `%load` that has nothing to do there I guess.

takluyver:
Yeah, the content of the docs needs some work (#13), but that will happen in another PR.

PR #19, opened 2015-06-30
remove some IPython-specific code

minrk:
- update a few genutils imports
- remove some special treatment of magics


takluyver:
This looks fine to me. We should remember this if someone comes complaining of strange regressions in tab completion in 4.0 around magics, though.

PR #20, opened 2015-06-30
use is_complete request to indicate complete input

minrk:
instead of IPython InputSplitter.

Adds a BlockingKernelClient to perform the is_complete_request, since the current use of `_is_complete` can't easily be made async.

based on #19

### DIFF
--- a/qtconsole/base_frontend_mixin.py
+++ b/qtconsole/base_frontend_mixin.py
@@ -1,5 +1,6 @@
-""" Defines a convenient mix-in class for implementing Qt frontends.
-"""
+"""Defines a convenient mix-in class for implementing Qt frontends."""
+
+from jupyter_client import BlockingKernelClient
 
 class BaseFrontendMixin(object):
     """ A mix-in class for implementing Qt frontends.
@@ -14,6 +15,7 @@ class BaseFrontendMixin(object):
     #---------------------------------------------------------------------------
     _kernel_client = None
     _kernel_manager = None
+    _blocking_client = None
 
     @property
     def kernel_client(self):
@@ -62,6 +64,23 @@ class BaseFrontendMixin(object):
         # we connected.
         if kernel_client.channels_running:
             self._started_channels()
+    
+    def _make_blocking_client(self):
+        kc = self.kernel_client
+        if kc is None:
+            return
+
+        info = kc.get_connection_info()
+        bc = BlockingKernelClient(**info)
+        bc.session.key = kc.session.key
+        bc.shell_channel.start()
+        self._blocking_client = bc
+    
+    @property
+    def blocking_client(self):
+        if self._blocking_client is None:
+            self._make_blocking_client()
+        return self._blocking_client
 
     @property
     def kernel_manager(self):
--- a/qtconsole/frontend_widget.py
+++ b/qtconsole/frontend_widget.py
@@ -8,17 +8,18 @@ from __future__ import print_function
 from collections import namedtuple
 import sys
 import uuid
+try:
+    from queue import Empty
+except ImportError:
+    from Queue import Empty
 
 from qtconsole import qt
 from qtconsole.qt import QtCore, QtGui
 from ipython_genutils import py3compat
 from ipython_genutils.importstring import import_item
 
-from IPython.core.inputsplitter import InputSplitter, IPythonInputSplitter
-from IPython.core.inputtransformer import classic_prompt
-from IPython.core.oinspect import call_tip
 from qtconsole.base_frontend_mixin import BaseFrontendMixin
-from traitlets import Any, Bool, Instance, Unicode, DottedObjectName
+from traitlets import Any, Bool, Float, Instance, Unicode, DottedObjectName
 from .bracket_matcher import BracketMatcher
 from .call_tip_widget import CallTipWidget
 from .history_console_widget import HistoryConsoleWidget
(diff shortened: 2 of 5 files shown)


qtconsole/base_frontend_mixin.py
```diff
@@ -62,6 +63,23 @@ def kernel_client(self, kernel_client):
         # we connected.
         if kernel_client.channels_running:
             self._started_channels()
+    
+    def _make_blocking_client(self):
+        kc = self.kernel_client
+        if kc is None:
+            return
+        from jupyter_client import BlockingKernelClient
```
takluyver: Should this be a regular import rather than a local one? jupyter_client is a plain dependency.

qtconsole/frontend_widget.py
```diff
@@ -233,14 +229,18 @@ def _is_complete(self, source, interactive):
...
         """
-        self._input_splitter.reset()
-        try:
-            complete = self._input_splitter.push(source)
-        except SyntaxError:
-            return True
-        if interactive:
-            complete = not self._input_splitter.push_accepts_more()
-        return complete
+        kc = self.blocking_client
```
takluyver: Your `.blocking_client` property has a code path that will return None. This code won't handle that.

qtconsole/frontend_widget.py
```diff
@@ -233,14 +229,18 @@ def _is_complete(self, source, interactive):
...
+        msg_id = kc.is_complete(source)
+        while True:
+            try:
+                reply = kc.shell_channel.get_msg(block=True, timeout=self.is_complete_timeout)
```
takluyver: Just to check my understanding, because this is a completely separate client, pulling these messages and discarding them if they're not the one we want won't influence the main asynchronous client that the Qt console is using, right?

qtconsole/frontend_widget.py
```diff
@@ -233,14 +229,18 @@ def _is_complete(self, source, interactive):
...
+            except Empty:
+                # assume incomplete output if we get no reply in time
+                return False
+            if reply['parent_header'].get('msg_id', None) == msg_id:
+                return reply['content']['status'] == 'complete'
+            else:
+                continue
```
takluyver: This else clause seems a bit redundant.

qtconsole/frontend_widget.py
```diff
@@ -344,7 +340,7 @@ def _insert_continuation_prompt(self, cursor):
         """ Reimplemented for auto-indentation.
         """
         super(FrontendWidget, self)._insert_continuation_prompt(cursor)
-        cursor.insertText(' ' * self._input_splitter.indent_spaces)
+        cursor.insertText(' ' * 4)
```
takluyver: The is_complete_reply does have an 'indent' field to provide a hint for this. I suspect this method is also called when you Ctrl+Enter to force a newline, so always inserting an indent is probably not right.

qtconsole/qtconsoleapp.py
```diff
@@ -205,6 +206,12 @@ def new_frontend_slave(self, current_widget):
         )
         kernel_client.load_connection_file()
         kernel_client.start_channels()
+        blocking_client = BlockingKernelClient(
+            connection_file=current_widget.kernel_client.connection_file,
+            config = self.config,
+        )
+        blocking_client.load_connection_file()
+        blocking_client.shell_channel.start()
```
takluyver: There's already a property to generate `blocking_client` - why is it being instantiated here? Won't it fail when you try to assign the attribute below, because the property is read only?

takluyver:
Thanks for getting round to this.

minrk:
Thanks for reviewing - I had this mostly put together when we last talked about it a few weeks ago, but forgot about it. Thanks for reviewing. I think I did things a few different ways, and there are likely a couple of redundant bits in here, still. I'll give it another pass tomorrow.

qtconsole/frontend_widget.py
minrk: yes, this just handles the case where previous replies arrived late due to timeout. The blocking client has no effect on messages received by the 'real' client.

qtconsole/frontend_widget.py
minrk: yup, the result of some refactoring.

qtconsole/frontend_widget.py
minrk: This now uses indent from the is_complete_reply

qtconsole/frontend_widget.py
minrk: handled now

qtconsole/base_frontend_mixin.py
minrk: regular, this was just lazy hacking as I was figuring it out.

qtconsole/qtconsoleapp.py
minrk: Yup, this one is totally unnecessary. Removed.

minrk:
Okay, review should be addressed. Some redundant bits from earlier attempts have been removed. indent from is_complete_reply is used now, instead of hardcoded value.

qtconsole/qtconsoleapp.py
```diff
@@ -57,6 +57,7 @@ def gui_excepthook(exctype, value, tb):
 )
 
 from jupyter_core.application import JupyterApp
+from jupyter_client import BlockingKernelClient
```
takluyver: This import is unnecessary

minrk: thanks, removed.

PR #24, opened 2015-07-02
remove the magics menu

minrk:
and other GUI controls that rely on IPython magics

This should remove any remaining dependency on IPython in the qtconsole.


Carreau:
+1

PR #26, opened 2015-07-03
remove remaining references to IPython

minrk:
- IPythonWidget is now JupyterWidget, etc.
- IPythonWidget config continues to work, with warnings
- IPythonWidget remains, with warnings


minrk:
based on #24

minrk:
Sorry, didn't notice this had conflicts. Should be ready to go, assuming Travis agrees.

takluyver:
Should we add subtstitutions to the config migration stuff in jupyter_core to change `[Rich]IPythonWidget` to `[Rich]JupyterWidget`? And then probably push out a quick release of that so it's in place before people start running the migration.

minrk:
Yes, I'll do that.

minrk:
jupyter/jupyter_core#36

PR #27, opened 2015-07-03
remove outdated references to IPython in docs

minrk:


docs/requirements.txt
```diff
@@ -1,6 +1,3 @@
 pygments
--e git+https://github.com/ipython/traitlets.git#egg=traitlets
--e git+https://github.com/jupyter/jupyter_core.git#egg=jupyter_core
--e git+https://github.com/jupyter/jupyter_client.git#egg=jupyter_client
```
takluyver: jupyter_client is not released yet (and this needs to keep a reference to direct dependencies)

minrk: Thanks, I put client back. Shouldn't the released dependencies get pulled in when this package is installed in the env?

takluyver: I think RTD runs `python setup.py install` instead of `pip install .`, so it won't use dependencies in our setup.py. Maybe we should make a PR to get it using pip. Or maybe we should just throw in the towel and accept that setuptools is here to stay. Though I think Fernando might fire us if we do that ;-)

PR #31, opened 2015-07-15
add basic install/run to README

minrk:


takluyver:
Installing with pip is only going to work if you have PyQt installed from another source, though. I'm not sure whether we should mention it, say what you've said, or skip over installation entirely.

minrk:
Probably mentioning that PyQt/PySide must come from somewhere else is a good idea.

damianavila:
>Probably mentioning that PyQt/PySide must come from somewhere else is a good idea.

Yep, I agree...

minrk:
Added mention of PyQt/PySide

damianavila:
:+1:

takluyver:
I think that was the last thing to do before releasing this. @minrk , do you want to push the release button?

minrk:
Yup, I'll hit it.

minrk:
Released. I think that leaves only:

- jupyter/notebook
- ipython/ipywidgets
- ipython/ipython
- jupyter meta pkg

to finish for 4.0

takluyver:
* New release of jupyter_core
* ipyparallel

minrk:
@takluyver I did both of those today, too.

takluyver:
Ninja releaser uses stealth!

PR #42, opened 2015-08-14
setuptools entrypoint

### LINKED ISSUE #43: launcher on windows
on windows, I get a "scripts\jupyter-qtconsole" instead of a hopefully "scripts\jupyter-qtconsole.exe"

(that I have for `jupyter-notebook.exe')

jupyter_launchers

using  qtconsole-4.0.0-py2.py3-none-any.whl
 

seems similar issue as there, but not in anaconda: https://groups.google.com/forum/#!topic/jupyter/ayLm0eivcBs

minrk:
needed for Windows

see jupyter/jupyter#13
closes #43

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -94,6 +94,12 @@ extras_require = setuptools_args['extras_require'] = {
 }
 
 if 'setuptools' in sys.modules:
+    setup_args['entry_points'] = {
+        'console_scripts': [
+            'jupyter-qtconsole = qtconsole.qtconsoleapp:main',
+        ]
+    }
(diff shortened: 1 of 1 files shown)


setup.py
```diff
@@ -94,6 +94,12 @@
 }
 
 if 'setuptools' in sys.modules:
+    setup_args['entry_points'] = {
+        'console_scripts': [
```
takluyver: I think this should have been gui_scripts, rather than console_scripts, to launch it without a console window on Windows.

PR #37, opened 2015-08-12
fix path for resources package_data

minrk:
closes #36

PR #46, opened 2015-08-17
Add qtconsole/resources to the package

flacjacket:
The JupyterConsole.svg icon was not included in the pypi package, this includes the relevant directory

### DIFF
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -2,6 +2,9 @@ include COPYING.md
 include CONTRIBUTING.md
 include README.md
 
+# Resources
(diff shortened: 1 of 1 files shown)


minrk:
#37 added it to package data, so it doesn't need to be in the manifest. This will be included in 4.0.1. Thanks!

flacjacket:
Ah, I didn't catch that (running on 4.0.0), thanks!

minrk:
I cut 4.0.1 just now, can you test that?

flacjacket:
Yeah, that fixed it, thanks again

PR #57, opened 2015-10-06
Removed LaTeX handling due to incomplete support

### LINKED ISSUE #56: LaTeX support is broken
In our [IRkernel](https://github.com/IRkernel/IRkernel), we use [repr](https://github.com/IRkernel/repr) to create mimebundles. this includes LaTeX and HTML for many things that aren’t plots.

This results in QtConsole displaying PNGs with LaTeX source code in them instead of something useful.

Please prefer the HTML version if available. Also LaTeX doesn’t necessarily mean “equation”.

flying-sheep:
Fixes the regression mentioned in #56

Please reject this PR by doing one fixing the functionality instead :smile:


Carreau:
Same than #64 slipped through the cracks. I think that even partial support is better than none.

PR #65, opened 2015-10-23
Fallback to the string printer if the latex printer raises an exception

### LINKED ISSUE #63: LaTeX that matplotlib fails on should fallback to the next renderer
There is a lot of advanced LaTeX that the matplotlib renderer cannot handle. If an object defines a LaTeX repr that it can't handle, it just shows the error message. It should fallback to the next renderer.

This is causing an issue for us with SymPy (see https://github.com/sympy/sympy/issues/9962). SymPy is able to print expressions using Unicode, but when the renderer fails (in this case, for matrices), it just shows the error 

```
Failed to render latex: 
\left[\begin{matrix}1 & 2\\3 & 4\end{matrix}\right]
      ^
Expected "\right" (at char 6), (line:1, col:7)
```

To reproduce:

- First, make sure you don't have `latex` installed in your PATH (or else it will just use that and not matplotlib)
- Install the latest version of SymPy
- Start the qtconsole and run

```
from sympy import *
init_session()
Matrix([[1, 2], [3, 4]])
```

asmeurer:
Currently it just prints an error and gives an empty png. However, there is a
lot of LaTeX that matplotlib cannot parse, e.g., matrices. This gives a better
user experience for SymPy, which has a text printer that it can fallback
to. SymPy cannot just disable the LaTeX printer because it is desired for the
notebook, and there is no way for it to enable LaTeX just in the notebook, or
change the printer precedence order.

I have left in the debug logging of the error (to the terminal).

Fixes #63. This supersedes #57, although I don't know if it fixes the issue
there.

### DIFF
--- a/qtconsole/rich_jupyter_widget.py
+++ b/qtconsole/rich_jupyter_widget.py
@@ -138,7 +138,12 @@ class RichJupyterWidget(RichIPythonWidget):
                 self._append_html(self.output_sep2, True)
             elif 'text/latex' in data:
                 self._pre_image_append(msg, prompt_number)
-                self._append_latex(data['text/latex'], True)
+                try:
+                    self._append_latex(data['text/latex'], True)
+                except Exception as e:
+                    self.log.error("Failed to render latex: '%s'", data['text/latex'], exc_info=True)
+                    self.log.error("Failed to render latex: %s" % e)
+                    return super(RichJupyterWidget, self)._handle_execute_result(msg)
                 self._append_html(self.output_sep2, True)
             else:
                 # Default back to the plain text representation.
@@ -166,7 +171,12 @@ class RichJupyterWidget(RichIPythonWidget):
                 jpg = decodestring(data['image/jpeg'].encode('ascii'))
                 self._append_jpg(jpg, True, metadata=metadata.get('image/jpeg', None))
             elif 'text/latex' in data and latex_to_png:
-                self._append_latex(data['text/latex'], True)
+                try:
+                    self._append_latex(data['text/latex'], True)
+                    self.log.error("Failed to render latex: '%s'", data['text/latex'], exc_info=True)
+                    self.log.error("Failed to render latex: %s" % e)
+                except Exception as e:
+                    return super(RichJupyterWidget, self)._handle_display_data(msg)
             else:
                 # Default back to the plain text representation.
                 return super(RichJupyterWidget, self)._handle_display_data(msg)
@@ -177,13 +187,7 @@ class RichJupyterWidget(RichIPythonWidget):
 
     def _append_latex(self, latex, before_prompt=False, metadata=None):
         """ Append latex data to the widget."""
-        try:
-            png = latex_to_png(latex, wrap=False)
-        except Exception as e:
-            self.log.error("Failed to render latex: '%s'", latex, exc_info=True)
-            self._append_plain_text("Failed to render latex: %s" % e, before_prompt)
-        else:
-            self._append_png(png, before_prompt, metadata)
+        self._append_png(latex_to_png(latex, wrap=False), before_prompt, metadata)
 
     def _append_jpg(self, jpg, before_prompt=False, metadata=None):
         """ Append raw JPG data to the widget."""
@@ -365,4 +369,3 @@ class RichIPythonWidget(RichJupyterWidget):
     def __init__(self, *a, **kw):
         warn("RichIPythonWidget is deprecated, use RichJupyterWidget")
         super(RichIPythonWidget, self).__init__(*a, **kw)
-
--- a/qtconsole/rich_jupyter_widget.py
+++ b/qtconsole/rich_jupyter_widget.py
@@ -172,8 +172,8 @@ class RichJupyterWidget(RichIPythonWidget):
             elif 'text/latex' in data and latex_to_png:
                 try:
                     self._append_latex(data['text/latex'], True)
-                    self.log.error("Failed to render latex: '%s'", data['text/latex'], exc_info=True)
                 except Exception:
+                    self.log.error("Failed to render latex: '%s'", data['text/latex'], exc_info=True)
                     return super(RichJupyterWidget, self)._handle_display_data(msg)
             else:
                 # Default back to the plain text representation.
--- a/qtconsole/rich_jupyter_widget.py
+++ b/qtconsole/rich_jupyter_widget.py
@@ -140,9 +140,8 @@ class RichJupyterWidget(RichIPythonWidget):
                 self._pre_image_append(msg, prompt_number)
                 try:
                     self._append_latex(data['text/latex'], True)
-                except Exception as e:
+                except Exception:
                     self.log.error("Failed to render latex: '%s'", data['text/latex'], exc_info=True)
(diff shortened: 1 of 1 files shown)


asmeurer:
@flying-sheep would love to hear if this fixes your issues or not. 

I didn't see any tests for latex, so didn't add any for this fix (because I'm not sure how).

flying-sheep:
hmm, no

asmeurer:
I'm unclear what exactly is happening in your case, then.

ccordoba12:
@flying-sheep, please clarify.

flying-sheep:
1. i installed the qtconsole dependencies in an almost clean venv (i symlinked my system pyqt)
2. i installed this branch of qtconsole
3. i started `jupyter qtconsole --kernel=ir`
4. i see what i show you in the screenshot

asmeurer:
@flying-sheep is any error information printed to the terminal? Maybe the problem in your case is that the latex rendering "fails" but doesn't register as a failure (raise an exception) to the qtconsole. If it did, it would fallback to the text renderer in this PR.

flying-sheep:
nothing.

Carreau:
Oops, slipped through te cracks with all the repos.

I'm fine with that.

flying-sheep:
excuse me: what is it that you’re fine with? :laughing:

Carreau:
> excuse me: what is it that you’re fine with?

The PR in general if it helps. The QtConsole is in low maintnance mode and few of the core dev use it.
So most PR should be merge accepted quickly. 

Frankly the QtConsole. should be replaced by a QtWebkitView that actually support a local mathjax install + codemirror/Ace.

asmeurer:
FYI the QTConsole does have users, as evidenced by the multiple bug reports we've received from SymPy for this bug. 

Also, in case you missed it, this PR and #57 are mutually exclusive with one another. You guys need to pick one or the other (or suggest a third alternative).

flying-sheep:
well, my problem as stated is that LaTeX → PNG isn’t a good solution to default to.

you can’t select & copy parts of the output, and it’s more overhead than simple text→text transformations. if the explicit plan is to replace the QtConsole frontend with a QWebView based one, i’m all for merging this and moving on, but if this is going to stay for a longer time, it’s pretty ugly for IRKernel users.

Carreau:
> FYI the QTConsole does have users, as evidenced by the multiple bug reports we've received from SymPy for this bug.

Yes I know. I'm just stating that it receives few patch from the core team. I would have no issues giving @flying-sheep , @asmeurer and @ccordoba12 commit rights if it makes things easier for you guys. (As long as a couple of other dev agrees also). You seem to be the one both using Qtconsole,  receiving bugs report,and writing patches, so it seem fair to me.

> Also, in case you missed it, this PR and #57 are mutually exclusive with one another. You guys need to pick one or the other (or suggest a third alternative).

Yes i though the incompatibility and why I commented on the other that `I think that even partial support is better than none.`

minrk:
I'm torn on this one. My current inclination is to go with this patch and allow explicit disable of latex via preferences.

ccordoba12:
@Carreau, I already have commit rights (@minrk granted them to me :-)

I'm +1 with @minrk proposal, I think it's a good compromise. I'll submit a PR with the option to disable Latex, if @flying-sheep and @asmeurer agree.

Please let me test and merge this one, I promise to do it this weekend ;-)

asmeurer:
That sounds good to me too.

(and btw I don't personally use the qtconsole; just fixing an issue for upstream SymPy)

asmeurer:
@ccordoba12 https://github.com/jupyter/qtconsole/issues/63 has steps to reproduce the SymPy issue for when you test this.

minrk:
@ccordoba12 go ahead and merge when you are ready.

flying-sheep:
well, disabling latex is already possible from the kernel, so two switches are probably confusing.

another idea: maybe we should hardcode a list of which kernels interpret “text/latex” like IPython does.

generally any MIME-type means of course everything compliant to the language/format specified by that mime type

but IPython means with “text/latex”: “a math subset of LaTeX that is roughly equivalent to what MathJax can do”

i don’t think we should accept all kernels to interpret the latex MIME type like IPython. it would be a hack but if IPython wants to keep doing that i think we should handle it in a special way.

ccordoba12:
@flying-sheep, how latex can be deactivated from the kernel?

flying-sheep:
```r
options(jupyter.plot_mimetypes = c('text/plain', 'image/png', 'image/svg+xml'))
```

qtconsole/rich_jupyter_widget.py
```diff
@@ -138,7 +138,12 @@ def _handle_execute_result(self, msg):
                 self._append_html(self.output_sep2, True)
             elif 'text/latex' in data:
                 self._pre_image_append(msg, prompt_number)
-                self._append_latex(data['text/latex'], True)
+                try:
+                    self._append_latex(data['text/latex'], True)
+                except Exception as e:
+                    self.log.error("Failed to render latex: '%s'", data['text/latex'], exc_info=True)
+                    self.log.error("Failed to render latex: %s" % e)
```
takluyver: I think the second log call is redundant here - `exc_info=True` in the first one should log the the error that was caught.

asmeurer: I pushed a fix for that.

asmeurer:
I just tested this, and it seems that the exception is printed in the qtconsole, which I believe wasn't the case when I originally made this. Any idea why that would be happening, and how to disable it? I only want the error to be logged to the terminal.

minrk:
@asmeurer are you sure the error isn't raised in the kernel? When I run the sample, sympy is computing the PNG kernel-side, so an error would logically be shown in the qtconsole.

asmeurer:
I have no idea what that even means, so ¯\\\_(ツ)\_/¯

minrk:
When I run the code (sympy 0.7.6.1):

```python
from sympy import *
init_session()
Matrix([[1, 2], [3, 4]])
```

sympy computes a png and latex string and sends them in a display message. The QtConsole receives this and displays the png computed by sympy. If sympy encountered an error during that computation, the traceback would be normal kernel output, and rightly displayed in the qtconsole.

asmeurer:
OK, my bad. That means that this is an issue in SymPy, and it turns out it has already been fixed (I was testing against the release, but I checked master and it went away).

qtconsole/rich_jupyter_widget.py
```diff
@@ -166,7 +170,11 @@ def _handle_display_data(self, msg):
                 jpg = decodestring(data['image/jpeg'].encode('ascii'))
                 self._append_jpg(jpg, True, metadata=metadata.get('image/jpeg', None))
             elif 'text/latex' in data and latex_to_png:
-                self._append_latex(data['text/latex'], True)
+                try:
+                    self._append_latex(data['text/latex'], True)
+                    self.log.error("Failed to render latex: '%s'", data['text/latex'], exc_info=True)
```
takluyver: This log call should be in the `except` block, not the `try` block, right?

asmeurer: Probably

PR #69, opened 2015-11-13
Updated example for creating inprocess qtconsole

lorton:
This was an example in the old IPython git repo, which was removed when
the code was moved here.  This is mostly just moved, but also changed
the imports to match qtconsole vs IPython.qt.*

### DIFF
--- a/examples/inprocess_qtconsole.py
+++ b/examples/inprocess_qtconsole.py
@@ -1,0 +1,37 @@
+"""
+An example of opening up an RichJupyterWidget in a PyQT Application, this can
+execute either stand-alone or by importing this file and calling
+inprocess_qtconsole.show().
+
+Based on the earlier example in the IPython repository, this has
+been updated to use qtconsole.
+"""
+
+
+from PyQt4 import QtGui
+from qtconsole.rich_jupyter_widget import RichJupyterWidget
+from qtconsole.inprocess import QtInProcessKernelManager
+
+
+def show():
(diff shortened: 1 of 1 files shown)


examples/inprocess_qtconsole.py
```diff
@@ -0,0 +1,37 @@
...
+An example of opening up an RichJupyterWidget in a PyQT Application, this can
+execute either stand-alone or by importing this file and calling
+inprocess_qtconsole.show().
+
+Based on the earlier example in the IPython repository, this has
+been updated to use qtconsole.
+"""
+
+
+from PyQt4 import QtGui
```
ccordoba12: You should import from `qtconsole` shim instead of doing a PyQt4 import. The thing is `qtconsole` works with PyQt4, PyQt5 and PySide.

ccordoba12:
The test failure is unrelated to your work, so please don't worry about it. I'll do a PR to fix that.

lorton:
OK - I've updated to use the qtconsole.qt.  I only have PyQt4 set up on my machines so I'm unfortunately not able to test the other ones, but it seems it should be OK.  Are the build tests run against different versions of PyQt/PySide?

ccordoba12:
> Are the build tests run against different versions of PyQt/PySide?

Nop, but they will be in a future. I asked you to do that change so that people can run the example with whatever bindings they have installed :-)

PR #71, opened 2015-11-24
Set AppUserModelID for Windows 7 and later

ccordoba12:
This way qtconsole uses its assigned taskbar icon instead of grabbing the one with the same AppUserModelID (which usually is the Python one)

----

@minrk, could it be possible to have a minor 4.1.1 release this week with this little change? This way we could include it in Anaconda 2.4.1, to be released next week. Thanks :-)


minrk:
Sure, no problem.

ccordoba12:
Thanks a lot! Please ping me when it's ready to update the Anaconda 
packages :-)

minrk:
Released 4.1.1 with this.

ccordoba12:
Thanks a lot @minrk!

PR #73, opened 2015-11-30
Provide an in-process version of RichJupyterWidget

### LINKED ISSUE #72: Multiline editing in an in-process console
When embedding a Jupyter console in a PyQt GUI using an in-process kernel, multiline editing is not triggered when a typed line is incomplete, although multiline editing works for pasted code. Instead, it raises a syntax error. The reason is because line 603 of console_widget.py calls the _is_complete function in frontend_widget.py, which attempts to send a message to a non-existent socket for a blocking kernel client and gets no reply. I believe that the solution is to subclass RichJupyterWidget (or possibly JupyterWidget), within inprocess.py, to provide an in-process version of '_is_complete', which checks for code completion directly. I will issue a pull request with a fix that I have tested.

rayosborn:
Fixes #72.
This is needed so that incomplete code typed into the console triggers multiline editing when using an in-process kernel.

### DIFF
--- a/qtconsole/inprocess.py
+++ b/qtconsole/inprocess.py
@@ -12,6 +12,7 @@ from .util import SuperQObject
 from .kernel_mixins import (
     QtKernelClientMixin, QtKernelManagerMixin,
 )
+from .rich_jupyter_widget import RichJupyterWidget
 
 class QtInProcessChannel(SuperQObject, InProcessChannel):
     # Emitted when the channel is started.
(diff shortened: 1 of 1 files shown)


takluyver:
Thanks! I think this is fine, but I'd like @minrk's opinion on it.

rayosborn:
I'm in no rush now. I can subclass the widget in my own code until it's released. This seems to work for a python console, but I have no idea whether it works for other languages.

takluyver:
I don't think we want to support using any other language in an in-process kernel. We're not even that keen on supporting it for Python, to be honest.

rayosborn:
Please don't ditch in-process kernels just yet. We might move to a remote kernel model eventually, but the in-process kernel works very well in our application.

takluyver:
I don't think we're about to ditch it,  but because it's not used in any of our code (and hard to automate testing), it's easy for us to break it unintentionally.

We probably need a better separation between the console UI and the machinery handling code execution, so it's easier to switch between in and out of process. But I don't think anyone's about to invest a lot of time in refactoring the Qt console.

PR #77, opened 2016-01-04
Ignore NoneType execute results

### LINKED ISSUE #68: RichJupyterWidget.execute() leads to exception
When trying to execute code before showing the `RichJupyterWidget`, an exception is thrown.  When I traced the code, it makes sense, as neither parent class of `FrontendWidget` has a `_handle_execute_reply` function, but there is a code path that will call `super(FrontendWidget, self)._handle_execute_reply`.  The traceback is below:

```
  File ".../qtconsole/base_frontend_mixin.py", line 163, in _dispatch
    handler(msg)
  File ".../qtconsole/jupyter_widget.py", line 182, in _handle_execute_reply
    super(JupyterWidget, self)._handle_execute_reply(msg)
  File ".../qtconsole/frontend_widget.py", line 496, in _handle_execute_reply
    super(FrontendWidget, self)._handle_execute_reply(msg)
AttributeError: 'super' object has no attribute '_handle_execute_reply'
```

I can try to find a suitable solution and submit a pull request if there's interest in it?

lorton:
Also added a clearer warning if the results have an unexpected info.type.

Closes #68

### DIFF
--- a/qtconsole/frontend_widget.py
+++ b/qtconsole/frontend_widget.py
@@ -134,7 +134,7 @@ class FrontendWidget(HistoryConsoleWidget, BaseFrontendMixin):
 
     confirm_restart = Bool(True, config=True,
         help="Whether to ask for user confirmation when restarting kernel")
-    
+
     lexer_class = DottedObjectName(config=True,
         help="The pygments lexer class to use."
     )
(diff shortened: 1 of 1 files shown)


lorton:
Sorry for the whitespace changes, my vi+python-mode automatically deletes trailing whitespaces.

minrk:
Merged without whitespace changes in 8b3f0a5

takluyver:
Thanks for digging into this, @lorton

Issue #82, opened 2016-01-08
[docs] Redirect to readthedocs


Carreau:
Now that qtconsole docs [works on readthedocs](http://qtconsole.readthedocs.org/) we should migrate the various links and replace the current pages on http://jupyter.org/qtconsole/stable/ with redirects.

willingc:
@Carreau Agreed.

I'm going to update the README to refer to readthedocs.

If someone with repo maintenance privileges can update the link next to the repo description to https://qtconsole.readthedocs.org/en/stable/ that would be great.

Carreau:
> If someone with repo maintenance privileges can update the link next to the repo description to https://qtconsole.readthedocs.org/en/stable/ that would be great.

Good point. Done.

willingc:
@Carreau Thanks for the repo link update.

Here's a PR #85 to update the README. I'll open another PR for any link changes in the source docs. Then I'll take a look at the gh-pages if needed.

PR #85, opened 2016-01-13
Update README

willingc:
Update for ReadTheDocs
- Add badges
- Expand description and add graphic
- Added subtitles
- Added resource section

PR #86, opened 2016-01-13
Update index of Sphinx docs

willingc:
Addresses #82 for Sphinx docs.

### DIFF
--- a/docs/source/index.rst
+++ b/docs/source/index.rst
@@ -1,8 +1,8 @@
 .. _qtconsole:
 
-========================
-A Qt Console for Jupyter
-========================
+==========================
+The Qt Console for Jupyter
+==========================
 
 :Release: |release|
 :Date: |today|
@@ -14,16 +14,16 @@ To start the Qt console::
 .. toctree::
    :maxdepth: 2
 
-   config_options
    installation
+   config_options
    changelog
 
 Overview
 ========
-Qt console is a very lightweight application that largely feels like a
+The Qt console is a very lightweight application that largely feels like a
 terminal, but provides a number of enhancements only possible in a GUI, such as
 inline figures, proper multiline editing with syntax highlighting, graphical
-calltips, and much more. Qt console can use any Jupyter kernel.
+calltips, and much more. The Qt console can use any Jupyter kernel.
 
 .. figure:: _images/qtconsole.png
     :width: 400px
@@ -49,7 +49,7 @@ navigation. This is not yet configurable.
 Inline graphics
 ===============
 
-One of the most exciting features of the QtConsole is embedded figures.
+One of the most exciting features of the Qt Console is embedded figures.
 You can plot with matplotlib in IPython, or the plotting library of choice
 in your kernel.
 
@@ -62,11 +62,11 @@ in your kernel.
 Saving and Printing
 ===================
 
-QtConsole has the ability to save your current session, as either HTML or
-XHTML. Your inlinke figures
-will be PNG in HTML, or inlined as SVG in XHTML. PNG images have the option to
-be either in an external folder, as in many browsers' "Webpage, Complete"
-option, or inlined as well, for a larger, but more portable file.
+The Qt Console has the ability to save your current session, as either HTML or
+XHTML. Your inline figures will be PNG in HTML, or inlined as SVG in XHTML.
+PNG images have the option to be either in an external folder, as in many
+browsers' "Webpage, Complete" option, or inlined as well, for a larger, but
+more portable file.
 
 .. note::
 
@@ -140,7 +140,7 @@ stylesheet:
 Fonts
 =====
 
-The QtConsole has configurable via the ConsoleWidget. To change these, set the
+The Qt console is configurable via the ConsoleWidget. To change these, set the
 ``font_family`` or ``font_size`` traits of the ConsoleWidget. For instance, to
 use 9pt Anonymous Pro::
 
@@ -375,10 +375,10 @@ Rules:
       the Kernel has been shutdown.
     * Remote Consoles may not restart or shutdown the kernel.
 
-Qt and the QtConsole
+Qt and the Qt console
 ====================
 
-An important part of working with the QtConsole when you are writing your own
+An important part of working with the Qt console when you are writing your own
 Qt code is to remember that user code (in the kernel) is *not* in the same
 process as the frontend.  This means that there is not necessarily any Qt code
 running in the kernel, and under most normal circumstances there isn't.
--- a/docs/source/index.rst
+++ b/docs/source/index.rst
@@ -9,19 +9,21 @@ A Qt Console for Jupyter
 
 To start the Qt console::
 
-    $> jupyter qtconsole
+    $ jupyter qtconsole
 
 .. toctree::
    :maxdepth: 2
 
    config_options
+   installation
    changelog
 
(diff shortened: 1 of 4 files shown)


docs/source/index.rst
```diff
@@ -75,6 +77,12 @@ or install it using your system package manager, e.g.::
 
     sudo apt-get install python3-pyqt5
 
+or::
+
+    sudo apt-get install python-pyqt
+    sudo apt-get install python3-pyqt
+
+if PyQt4 is desired.
```
takluyver: I'm not sure it's necessary to illustrate this; there's no particular reason the end user should care about using Qt 4 vs Qt 5. If they're using the Qt console as part of another Qt 4 application, the instructions for that application should get them set up.

If we do want to show it, then the names are `python-qt4` and `python3-pyqt4`.

willingc: @takluyver Thanks for the correction. I don't feel strongly about keeping in or taking out. Since the code base still mentions Qt 4, it might be good to just leave it in but corrected.

ccordoba12: @takluyver, I'm in favor of keeping the pyqt4 instructions (and sorry for the name errors, I should have verified them first).

The thing is only leaving `python3-pyqt5` seems to imply that `qtconsole` only works in Python 3 and with PyQt5, which is obviously not true.

takluyver: Maybe in the paragraph above, we should expand "PyQt or PySide" to say "PyQt5, PyQt4 or PySide".

I don't think there's any great harm in putting these in, but it seems like superfluous information. There's a lot of potential combinations of packaging system, Python version and Qt binding, and I don't want to end up having to make a great big list of possible commands. I'd rather give one example command and let people work out how to adapt it.

ccordoba12: I propose to leave the installation instructions for PyQt4 because (as you said) there's no harm in having them, but *don't* add further instructions for other Linux distributions and Qt bindings.

takluyver: The reason I'd rather not put them in is that once we have more than one example, there's no obvious reason why we're not adding others. If PyQt4 is worth specific mention, why not PySide? Why not commands for Fedora, or Arch Linux?

I also think that when we write docs, we have a tendency to leave in extra information because of a vague feeling that 'more information never hurts', and that gradually leads to longwinded docs that no-one can be bothered to read. So I try to resist that.

ccordoba12: Then we should remove the reference to Ubuntu and `python3-pyqt5` as well because (as I said) it's misleading: casual users will believe that `qtconsole` only works on Python 3 and PyQt5.

For what is worth, we don't provide specific installation instructions for any Linux distribution in Spyder, so I'm fine with going with that too here :-)

takluyver: How about if, in the paragraph above, we expand it to say "PyQt5, PyQt4 or PySide". The install line is introduced with "e.g.", which means that it's an example, so that seems pretty clear to me.

ccordoba12: Ok, fair enough :-) But then I'd like to see something really explicit, like

> e.g. to install PyQt5 on Ubuntu and Python 3 you need to run
> `sudo apt-get install python3-pyqt5`

willingc: I'm going to rework this based on both of your concerns (verbosity/too much info and clarity/supporting the casual users).

takluyver: Thanks @willingc !

ccordoba12: :+1:

willingc:
@takluyver @ccordoba12 I've now updated the Sphinx docs for installation as well based on the content in PR #85. I broke out installation into its own rst file. In the next few weeks, I plan to loop back break up the index.rst to usage.rst and security.rst as well.

docs/source/index.rst
```diff
@@ -9,19 +9,21 @@ A Qt Console for Jupyter
...
 To start the Qt console::
 
-    $> jupyter qtconsole
+    $ jupyter qtconsole
 
 .. toctree::
    :maxdepth: 2
 
    config_options
+   installation
```
takluyver: I'd put installation before config options

docs/source/index.rst
```diff
@@ -9,19 +9,21 @@ A Qt Console for Jupyter
...
    changelog
 
-You can use any Jupyter kernel with this PyQt_ console GUI.
-This is a very lightweight widget that
-largely feels like a terminal, but provides a number of enhancements only
-possible in a GUI, such as inline figures, proper multiline editing with syntax
-highlighting, graphical calltips, and much more.
+Overview
+========
+Qt console is a very lightweight application that largely feels like a
```
takluyver: We tend to say "The Qt console", rather than just "Qt console"

docs/source/installation.rst
```diff
@@ -0,0 +1,62 @@
+Installation
+============
+The qtconsole requires Qt, such as
```
takluyver: For consistency with other parts of the docs, let's say "The Qt console" rather than "The qtconsole".

As I see it:
* "The Qt console" is what we call the application
* "qtconsole" is the command that starts the application

takluyver:
A couple of minor points about wording, but overall this looks good.

docs/source/index.rst
willingc: Got it. Thanks for clarifying. I will go back over @takluyver :+1:

docs/source/index.rst
willingc: +1

willingc:
@takluyver I've made your suggested changes.

docs/source/installation.rst
```diff
@@ -0,0 +1,62 @@
...
+
+   sudo apt-get install python3-pyqt5
+
+to install PyQt5,::
+
+   sudo apt-get install python-pyqt
+
+or::
+
+   sudo apt-get install python3-pyqt
```
takluyver: They're still not called this! It's `python3-pyqt4` and `python-qt4`. ;-)

You can search on http://packages.ubuntu.com/ to check any package names.

docs/source/installation.rst
```diff
@@ -0,0 +1,62 @@
...
+    conda install pyqt
+
+or with a system package manager. For Windows, PyQt binary packages may be
+used.
+
+For example with Linux Debian's system package manager, use::
+
+   sudo apt-get install python3-pyqt5
+
+to install PyQt5,::
```
takluyver: For people scan-reading, I think it's counterintuitive to have the descriptions after the things they refer to - at a glance, this will look like the *next* example is for PyQt5.

Maybe condense the examples into one block with comments?

```
sudo apt-get install python3-pyqt5  # PyQt5 on Python 3
sudo apt-get install python3-pyqt4  # PyQt4 on Python 3
sudo apt-get install python-qt4     # PyQt4 on Python 2
```

docs/source/installation.rst
willingc: My bad. Good catch. Fixing...

docs/source/installation.rst
willingc: Sure. I can do the block. It always kind of bugs me to have a comment on something that would be typed on the command line. But since this is an example...I will change ;)

Carreau:
Thanks !

willingc:
:city_sunset: @Carreau @takluyver @ccordoba12

PR #88, opened 2016-01-19
If a Qt binding is already loaded, use it.

### LINKED ISSUE #87: Override QT_API when a Qt binding is already loaded?
I would like to suggest not obeying QT_API when a specific Qt binding has already been loaded, and just use it, rather than raising an exception: the error is probably just a misconfigured environment variable and there's a clear, single, well-defined recovery.

anntzer:
Closes #87.

### DIFF
--- a/qtconsole/qt_loaders.py
+++ b/qtconsole/qt_loaders.py
@@ -240,36 +240,41 @@ def import_pyside():
 
 def load_qt(api_options):
     """
-    Attempt to import Qt, given a preference list
-    of permissible bindings
+    Import Qt, given a preference list of permissible bindings.
 
-    It is safe to call this function multiple times.
+    It is safe to call this function multiple times.  If a Qt binding
+    has already been loaded, it will be returned regardless of the
+    preference list.
 
     Parameters
     ----------
     api_options: List of strings
-        The order of APIs to try. Valid items are 'pyside',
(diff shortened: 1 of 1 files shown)


qtconsole/qt_loaders.py
```diff
@@ -240,30 +240,37 @@ def import_pyside():
...
 
     Raises
     ------
-    ImportError, if it isn't possible to import any requested
-    bindings (either becaues they aren't installed, or because
-    an incompatible library has already been installed)
+    ImportError, if it isn't possible to import any requested bindings
+    (because they aren't installed).
     """
+    if "PySide" in sys.modules:
```
takluyver: This should use the `loaded_api()` check that already exists in the module. It's possible, if unusual, to load the top-level namespace of one of these packages without actually loading the bindings:

```
In [1]: import PyQt4

In [2]: import PyQt5.QtCore

In [3]: 
```

`loaded_api()` checks whether `QtCore` is loaded.

anntzer: fixed.

takluyver:
Thanks, looks good to me. I'll give other people a day or so to look at it, then merge.

minrk:
:+1: @ccordoba12?

ccordoba12:
:+1: from me too

PR #92, opened 2016-01-23
Fix number of columns for ncurses tab completion

### LINKED ISSUE #84: Output more possibilities on tab-completion
Currently, tab completion only displays a small list of possibilities at a time:
completion
The list is also always displayed in three columns, even if they don't fit (then the last column is just invisible) or if the window is much wider than that.
Given that the qtconsole can erase the unused completions I'd suggest just displaying the whole list instead (which is also useful as a quick substitute to calling `dir`/`vars`).
(Actually, a pager-like behavior would be more useful in the console ipython, which currently outputs the whole list without erasing it afterwards.)

vidartf:
Previously the calculation for how to lay out the tab-completion suggestions just assumed a display width of 80 characters. This fixes it to use the correct width.

Fixes #84.

### DIFF
--- a/qtconsole/completion_html.py
+++ b/qtconsole/completion_html.py
@@ -37,7 +37,9 @@ def html_tableify(item_matrix, select=None, header=None , footer=None) :
         foot = (u'<tr>'\
             +''.join((u'<td>'+footer+u'</td>')*len(item_matrix[0]))\
             +'</tr>')
-    html = (u'<table class="completion" style="white-space:pre">'+head+(u''.join(html_cols))+foot+u'</table>')
+    html = (u'<table class="completion" style="white-space:pre"'
+            'cellspacing=0>' +
+            head + (u''.join(html_cols)) + foot + u'</table>')
     return html
 
 class SlidingInterval(object): 
--- a/qtconsole/completion_html.py
+++ b/qtconsole/completion_html.py
@@ -302,7 +302,12 @@ class CompletionHtml(QtGui.QWidget):
             return
         self._start_position = cursor.position()
         self._consecutive_tab = 1
-        items_m, ci = text.compute_item_matrix(items, empty=' ')
+        # Calculate the number of characters available.
+        width = self._text_edit.document().textWidth()
+        char_width = QtGui.QFontMetrics(self._console_widget.font).width(' ')
+        displaywidth = int(max(10, (width / char_width) - 1))
+        items_m, ci = text.compute_item_matrix(items, empty=' ',
+                                               displaywidth=displaywidth)
         self._sliding_interval = SlidingInterval(len(items_m)-1)
 
         self._items = items_m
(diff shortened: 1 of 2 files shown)


vidartf:
PS: I did some style fixes as well, so please consider checking the commits for a cleaner diff.

takluyver:
We don't really do style cleanups - we treat PEP8 very much as a loose guideline, not a hard and fast rule. Can you rebase the branch to eliminate that commit, so the PR is just the functional change.

ccordoba12:
I agree with @takluyver, Your pep8 changes make hard to review your PR.

vidartf:
Rebased and force pushed.

takluyver:
I think this looks good, but, I'll let people more familiar with this look into it.

qtconsole/completion_html.py
```diff
@@ -302,7 +304,12 @@ def show_items(self, cursor, items):
             return
         self._start_position = cursor.position()
         self._consecutive_tab = 1
-        items_m, ci = text.compute_item_matrix(items, empty=' ')
+        # Calculate the number of characters available.
+        width = self._text_edit.document().textWidth()
+        char_width = QtGui.QFontMetrics(self._console_widget.font).width(' ')
+        displaywidth = int(max(10, (width / char_width) - 1))
```
ccordoba12: Could you explain what's the purpose of `displaywidth`?

vidartf: The width of the terminal in number of characters. In other words: How many characters that can fit into each line in the QTextEdit (i.e. the terminal). The default `displaywidth` used in `text.compute_item_matrix()` is 80, which is why it previously disregarded widget size when laying out the completion suggestions.

ccordoba12: So here you're making the minimal width to be 10 chars, which I think is too small. Why not 80?

takluyver: People can reasonably use a console <80 characters wide, so I think the minimum should be less than 80. I don't have strong feelings within the range of about 10-40, though.

vidartf: This code is simply copied from `ConsoleWidget._format_as_columns()` (which I adjusted prior to that). I thought it best to follow whatever conventions was already in place.

ccordoba12: We could take the opportunity to change it. I'm in favor of using 40 :-)

vidartf: PS: This code should in principle do exactly the same as `ConsoleWidget._format_as_columns()`, but it needs the raw matrix as it needs to have the grid for highlighting and selection changes, as well as for displaying in HTML.

ccordoba12: Ok, let's not change things in this PR because they don't belong to the problem you're trying to solve.

ccordoba12:
This is working correctly, so merging :-)

PR #93, opened 2016-01-23
Fix droplist completer focus

vidartf:
Currently, if you click on the `CompletionWidget` with the mouse, it gets focus (Windows 10, 64-bit, PyQt4). This triggers the `CompletionWidget` to hide. To prevent this, check the active window before hiding. Next, since it now has focus, any keyboard input to the `CompletionWidget` needs to be redirected to its `_text_editor`, except the input that is intended for it (up/down/page up/page down/home/end).

### DIFF
--- a/qtconsole/completion_widget.py
+++ b/qtconsole/completion_widget.py
@@ -44,7 +44,7 @@ class CompletionWidget(QtGui.QListWidget):
         self.itemActivated.connect(self._complete_current)
 
     def eventFilter(self, obj, event):
-        """ Reimplemented to handle keyboard input and to auto-hide when the
+        """ Reimplemented to handle mouse input and to auto-hide when the
             text edit loses focus.
         """
         if obj is self:
@@ -87,7 +87,6 @@ class CompletionWidget(QtGui.QListWidget):
         """
         super(CompletionWidget, self).hideEvent(event)
         self._text_edit.cursorPositionChanged.disconnect(self._update_current)
-        self._text_edit.removeEventFilter(self)
         self.removeEventFilter(self)
 
     def showEvent(self, event):
@@ -95,7 +94,6 @@ class CompletionWidget(QtGui.QListWidget):
         """
         super(CompletionWidget, self).showEvent(event)
         self._text_edit.cursorPositionChanged.connect(self._update_current)
-        self._text_edit.installEventFilter(self)
         self.installEventFilter(self)
 
     #--------------------------------------------------------------------------
--- a/qtconsole/completion_widget.py
+++ b/qtconsole/completion_widget.py
@@ -136,6 +136,7 @@ class CompletionWidget(QtGui.QListWidget):
         self.setGeometry(point.x(), point.y(), w, height)
         self._start_position = cursor.position()
         self.setCurrentRow(0)
+        self.raise_()
         self.show()
 
     #--------------------------------------------------------------------------
--- a/qtconsole/completion_widget.py
+++ b/qtconsole/completion_widget.py
@@ -95,17 +95,18 @@ class CompletionWidget(QtGui.QListWidget):
         text_edit = self._text_edit
         point = text_edit.cursorRect(cursor).bottomRight()
         point = text_edit.mapToGlobal(point)
-        height = self.sizeHint().height()
-        screen_rect = QtGui.QApplication.desktop().availableGeometry(self)
-        if (screen_rect.size().height() + screen_rect.y() 
-                    - point.y() - height < 0):
-            point = text_edit.mapToGlobal(text_edit.cursorRect().topRight())
-            point.setY(point.y() - height)
-        self.move(point)
-
-        self._start_position = cursor.position()
         self.clear()
         self.addItems(items)
+        height = self.sizeHint().height()
+        screen_rect = QtGui.QApplication.desktop().availableGeometry(self)
+        if (screen_rect.size().height() + screen_rect.y() -
+                point.y() - height < 0):
+            point = text_edit.mapToGlobal(text_edit.cursorRect().topRight())
+            point.setY(point.y() - height)
+        w = (self.sizeHintForColumn(0) +
+             self.verticalScrollBar().sizeHint().width())
+        self.setGeometry(point.x(), point.y(), w, height)
+        self._start_position = cursor.position()
         self.setCurrentRow(0)
         self.show()
 
--- a/qtconsole/completion_widget.py
+++ b/qtconsole/completion_widget.py
@@ -68,8 +68,7 @@ class CompletionWidget(QtGui.QListWidget):
                 pos = self.mapToGlobal(event.pos())
                 target = QtGui.QApplication.widgetAt(pos)
                 if (target and self.isAncestorOf(target) or target is self):
-                    target.event(event)
-                    return True
+                    return False
                 else:
                     self.cancel_completion()
                     if target:
--- a/qtconsole/completion_widget.py
+++ b/qtconsole/completion_widget.py
@@ -40,7 +40,7 @@ class CompletionWidget(QtGui.QListWidget):
             etype = event.type()
 
             if etype == QtCore.QEvent.KeyPress:
-                key, text = event.key(), event.text()
+                key = event.key()
                 if key in (QtCore.Qt.Key_Return, QtCore.Qt.Key_Enter,
                            QtCore.Qt.Key_Tab):
                     self._complete_current()
--- a/qtconsole/completion_widget.py
+++ b/qtconsole/completion_widget.py
@@ -39,10 +39,10 @@ class CompletionWidget(QtGui.QListWidget):
             flags = QtCore.Qt.ToolTip | QtCore.Qt.WindowStaysOnTopHint
 
         self.setAttribute(QtCore.Qt.WA_StaticContents)
-        origPolicy = text_edit.focusPolicy()
+        original_policy = text_edit.focusPolicy()
         self.setWindowFlags(flags)
         self.setFocusPolicy(QtCore.Qt.NoFocus)
-        text_edit.setFocusPolicy(origPolicy)
+        text_edit.setFocusPolicy(original_policy)
 
         # Ensure that the text edit keeps focus when widget is displayed.
         self.setFocusProxy(self._text_edit)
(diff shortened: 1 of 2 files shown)


vidartf:
This should also fix spyder-ide/spyder#1908 and spyder-ide/spyder#2265.

ccordoba12:
I've found a problem with your pull request. Steps to reproduce:

1. Write `import <Tab>` and select an entry in the completion widget with a **single** mouse click.
2. Start to write some text right after doing a selection, e.g. `set`
3. Do a **single** mouse click out of qtconsole (e.g. in Chrome or Firefox).
4. You can see the completion widget is not hidden (as it should be) but instead continues to be above qtconsole.

vidartf:
Fixed. Rebased, so anything based off the last one is detached. Now does some magic ritual in the initialization that I'm not 100% sure of what does, but I can hazard a guess. Now it also sets the width of the widget correctly (and therefore forces horz scrollbar off).

vidartf:
This now works smoothly on Windows, but it would be very helpful if someone on other systems could verify.

takluyver:
Just tested on Linux (anyone else testing, the option you need is `--gui-completion=droplist`). Unfortunately, with this PR, the dropdown for completions doesn't appear at all. It's still making the completion request, and if there's only one possible completion, it fills it in, but if there's more than one, no dice.

No error messages when this happens.

vidartf:
Would you mind playing with `git reset` to see at which commit this start happening?

takluyver:
I used git bisect, and it say 8917eeb0683e9047f7dc55051c65cdae0f6ee7ed is where it went wrong:

```
8917eeb0683e9047f7dc55051c65cdae0f6ee7ed is the first bad commit
commit 8917eeb0683e9047f7dc55051c65cdae0f6ee7ed
Author: Vidar Tonaas Fauske <vidartf@gmail.com>
Date:   Sun Jan 24 04:30:56 2016 +0100

    Add some CompletionWidget setup magic
```

vidartf:
PyQt or PySide?

takluyver:
Looks like it's picking up PyQt5 on my system.

vidartf:
Could you try to run the tests I checked in and see which i any succeed?

takluyver:
They all pass (I see another failure, but I see the same in master, and I think it's unrelated).

vidartf:
Could you try re-adding the  `QtCore.Qt.WindowStaysOnTopHint` window flag?

takluyver:
I still don't see the dropdown.

vidartf:
Well, as far as I can understand, it should be the code in `__init__` that caused the change. Of that code it could either be the window flags (popup vs tooltip), focus policy, or the four lines that were added to the start (or a combination of these). If you'd be able to revert thing piece by piece, we could maybe figure out why it doesn't work.

takluyver:
It's the window flags - if I change `Popup` back to `ToolTip`, it works.

vidartf:
Could you try adding the following statement somewhere in the init?
```python
self.setAttribute(QtCore.Qt.WA_X11NetWmWindowTypeCombo)
```

vidartf:
... and/or adding this to the end of `show_items()` just before `self.show()`: `self.raise_()` (since all the tests seems to indicate that it is shown, it might just have the wrong Z-order?).

vidartf:
If neither of those work, we could try to add a platform conditional for which flag(s) to use? Is there a standard way for this project for how to check the platform?

ccordoba12:
> Is there a standard way for this project for how to check the platform?

Python has it:

1. `os.name == 'nt'`: Windows
2. `sys.platform == 'darwin'`: OS X
3. `sys.platform.startswith('linux')`: Linux

takluyver:
Neither of those did the trick, either singly or in combination.

Thanks for your persistence with this!

vidartf:
Ok, added platform specific flags. The PR should still fix the issue of horizontal scroller / popup width as discussed in spyder-ide/spyder#2265. Care to confirm that everything works as advertised?

Note: I assumed the previous behavior worked on any OS other than Windows. If that is not the case, it should be easy to tune.

takluyver:
That seems to be working for me. Thanks!

qtconsole/completion_widget.py
```diff
@@ -20,9 +21,21 @@ def __init__(self, console_widget):
         super(CompletionWidget, self).__init__()
 
         self._text_edit = text_edit
+        self.setEditTriggers(QtGui.QAbstractItemView.NoEditTriggers)
+        self.setHorizontalScrollBarPolicy(QtCore.Qt.ScrollBarAlwaysOff)
+        self.setSelectionBehavior(QtGui.QAbstractItemView.SelectRows)
+        self.setSelectionMode(QtGui.QAbstractItemView.SingleSelection)
+
+        if os.name == 'nt':
```
takluyver: Let's leave a brief comment here explaining why we need different flags for different systems.

We should also try to get someone with OSX to check that it's working for them.

vidartf: Pushed comments. If there's someone on OSX that can test this in time for the appropriate milestone, then all the better, if not, it does at least fix an existing issue for Windows users.

qtconsole/completion_widget.py
```diff
@@ -20,9 +21,28 @@ def __init__(self, console_widget):
...
+            # ToolTip style)
+            flags = QtCore.Qt.Popup
+        else:
+            # On other OS, we need ToolTip style to ensure that the widget
+            # shows correctly (dialog is not drawn with Popup style)
+            flags = QtCore.Qt.ToolTip | QtCore.Qt.WindowStaysOnTopHint
 
         self.setAttribute(QtCore.Qt.WA_StaticContents)
-        self.setWindowFlags(QtCore.Qt.ToolTip | QtCore.Qt.WindowStaysOnTopHint)
+        origPolicy = text_edit.focusPolicy()
```
ccordoba12: Please change this to `original_policy` to follow the Python custom of naming variables with underscores instead of camelcase.

qtconsole/tests/test_completion_widget.py
```diff
@@ -0,0 +1,70 @@
...
+from qtconsole.completion_widget import CompletionWidget
+import ipython_genutils.testing.decorators as dec
+
+api = loaded_api()
+if api in (QT_API_PYQT, QT_API_PYQTv1, QT_API_PYQT_DEFAULT):
+    from PyQt4.QtTest import QTest
+elif api == QT_API_PYQT5:
+    from PyQt5.QtTest import QTest
+elif api == QT_API_PYSIDE:
+    from PySide.QtTest import QTest
```
ccordoba12: I think this could be simplified by only importing `QT_API_PYQT5` and `QT_API_PYSIDE` and writing this block as

```python
if api == QT_API_PYQT5:
    from PyQt5.QtTest import QTest
elif api == QT_API_PYSIDE:
    from PySide.QtTest import QTest
else:
    from PyQt4.QtTest import QTest
```

ccordoba12:
@vidartf, I'm sorry to say it but your changes doesn't work for me on Linux.

If I select an entry in the completion widget with a mouse right-click, it hides immediately, just as before.

ccordoba12:
Things were working when I tested your branch last week though :-s

ccordoba12:
I should have tested further: the failure I reported only appears on PyQt5. On PyQt4 your changes work as expected.

vidartf:
@ccordoba12, on PyQt5, could you test with using the `Qt.Popup` flags? Maybe only linux + PyQt4 needs `Qt.ToolTip`?

ccordoba12:
@vidartf, that's not the problem. I used `Qt.Popup` instead of `Qt.ToolTip`and the widget doesn't show up (as reported by @takluyver).

vidartf:
Does linux + PyQt5 work with the old code?

ccordoba12:
Not, it doesn't. The completion widget in not displayed with your changes from 8 days ago.

vidartf:
Not sure if we talked past each other or not. I meant: Does droplist work with linux + PyQt5 on master?

ccordoba12:
> Does droplist work with linux + PyQt5 on master?

Yes, the completion widget is shown as expected.

vidartf:
> Yes, the completion widget is shown as expected.

And can be clicked?

ccordoba12:
Nop, when you click on any entry the widget disappears immediately (but 
that is also the case with PyQt4 on master).

vidartf:
Oh. From spyder-ide/spyder#1908 I got the impression this was a Windows specific issue. Umm.. In that case I'm kind of out of ideas, except that you switch to `Popup` and try to figure out what is needed to get it to show on Linux? Maybe set the parent? Currently the completer doesn't have a parent (from the git log, somebody complained about segfaults on closing), so you might try setting that?

If you try the bottom-up approach, maybe test that you can get a generic QWidget/QWindow show with the `Popup` flag set, then switch to a `QListItemView`, and then progressively get closer to the flags/attributes of the completion widget. Or you could do top-down and see what needs to be removed for it to show.

Either way, it's kind of hard for me to debug as I don't currently have access to a linux machine.

ccordoba12:
Could you point to the commit that removed parent?

vidartf:
https://github.com/ipython/ipython/commit/f21a128a37a6f32ed502e2fe5bf4e82a35875043

ccordoba12:
@vidartf, I sent a pull request against your branch that fixes the issues I'm seeing with PyQt5.

Please test it on Windows to see if it didn't break anything there :-)

vidartf:
I've merged the proposed changed from @ccordoba12. Would you mind confirming whether this still works on your setup @takluyver ?

takluyver:
That appears to be working for me with PyQt5.

vidartf:
Alright, I cleaned up some more code, and think this can be reviewed for merge again. If possible, it should be tested with PySide and on Mac OSX.

takluyver:
Checked under PySide + Python 2 on Linux (I was testing with Python 3 before).

@Carreau have you got a few minutes to check this PR on a Mac and check that it hasn't broken anything? You'll need to run with `--gui-completion=droplist` to test it.

Carreau:
> have you got a few minutes to check this PR on a Mac and check that it hasn't broken anything? You'll need to run with --gui-completion=droplist to test it.

Seem to work for me.

ccordoba12:
Ok, everything seems to be working fine for me too, on both PyQt4 and PyQt5 with Linux.

Thanks a lot for your work on this @vidartf!

PR #95, opened 2016-02-02
Fix Latex support

### LINKED ISSUE #56: LaTeX support is broken

ccordoba12:
Fixes #56 

This builds on top of ipython/ipython#9203, so that one has to be merged first.

Pinging @flying-sheep, @asmeurer and @jorisvandenbossche

----

Tasks:

- [x] Fix the case when neither Matplotlib nor Latex are installed. Qtconsole was displaying an empty image when it should move to use the plain text repr.
- [x] Fix the case when Matplotlib is incapable of handling Latex. Qtconsole was printing a huge traceback, even after pull request #65 .
- [x] Fix the case when Latex doesn't come in math mode. This will fix the issues with IRKernel and Pandas.
- [x] Bonus: First try to use **Latex**, then **Matplotlib** to generate the representation.

### DIFF
--- a/qtconsole/rich_jupyter_widget.py
+++ b/qtconsole/rich_jupyter_widget.py
@@ -138,11 +138,11 @@ class RichJupyterWidget(RichIPythonWidget):
                 self._append_html(self.output_sep2, True)
             elif 'text/latex' in data:
                 self._pre_image_append(msg, prompt_number)
-                try:
-                    self._append_latex(data['text/latex'], True)
-                except Exception:
-                    self.log.error("Failed to render latex: '%s'", data['text/latex'], exc_info=True)
-                    return super(RichJupyterWidget, self)._handle_execute_result(msg)
+                png = latex_to_png(data['text/latex'], wrap=False)
+                if png:
+                    self._append_png(png, True)
+                else:
+                    return super(RichJupyterWidget, self)._handle_display_data(msg)
                 self._append_html(self.output_sep2, True)
             else:
                 # Default back to the plain text representation.
@@ -170,10 +170,10 @@ class RichJupyterWidget(RichIPythonWidget):
                 jpg = decodestring(data['image/jpeg'].encode('ascii'))
                 self._append_jpg(jpg, True, metadata=metadata.get('image/jpeg', None))
             elif 'text/latex' in data and latex_to_png:
-                try:
-                    self._append_latex(data['text/latex'], True)
-                except Exception:
-                    self.log.error("Failed to render latex: '%s'", data['text/latex'], exc_info=True)
+                png = latex_to_png(data['text/latex'], wrap=False)
+                if png:
+                    self._append_png(png, True)
+                else:
                     return super(RichJupyterWidget, self)._handle_display_data(msg)
             else:
                 # Default back to the plain text representation.
@@ -182,10 +182,6 @@ class RichJupyterWidget(RichIPythonWidget):
     #---------------------------------------------------------------------------
     # 'RichJupyterWidget' protected interface
     #---------------------------------------------------------------------------
-
-    def _append_latex(self, latex, before_prompt=False, metadata=None):
-        """ Append latex data to the widget."""
-        self._append_png(latex_to_png(latex, wrap=False), before_prompt, metadata)
 
     def _append_jpg(self, jpg, before_prompt=False, metadata=None):
         """ Append raw JPG data to the widget."""
--- a/qtconsole/rich_jupyter_widget.py
+++ b/qtconsole/rich_jupyter_widget.py
@@ -183,9 +183,13 @@ class RichJupyterWidget(RichIPythonWidget):
 
     def _append_latex(self, latex, before_prompt=False, metadata=None):
         """ Append latex data to the widget."""
-        png = latex_to_png(latex, wrap=False)
-        if png:
-            self._append_png(png, before_prompt, metadata)
+        supported_latex = latex.startswith('$') and latex.endswith('$')
+        if supported_latex:
+            png = latex_to_png(latex, wrap=False)
+            if png:
+                self._append_png(png, before_prompt, metadata)
+            else:
+                raise ValueError
         else:
             raise ValueError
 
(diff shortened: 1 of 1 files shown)


qtconsole/rich_jupyter_widget.py
```diff
@@ -140,9 +140,8 @@ def _handle_execute_result(self, msg):
                 self._pre_image_append(msg, prompt_number)
                 try:
                     self._append_latex(data['text/latex'], True)
-                except Exception:
-                    self.log.error("Failed to render latex: '%s'", data['text/latex'], exc_info=True)
```
minrk: Isn't it a *good* thing to log failure to render the latex?

qtconsole/rich_jupyter_widget.py
```diff
@@ -185,7 +183,15 @@ def _handle_display_data(self, msg):
 
     def _append_latex(self, latex, before_prompt=False, metadata=None):
         """ Append latex data to the widget."""
-        self._append_png(latex_to_png(latex, wrap=False), before_prompt, metadata)
+        supported_latex = latex.startswith('$') and latex.endswith('$')
```
flying-sheep: this is missing the other mathmode-triggering syntaxes: https://github.com/jupyter/qtconsole/issues/56#issuecomment-178167408

we should add a more sophisticated check extracted to a utility function, e.g. called `is_latex_mathmode`

qtconsole/rich_jupyter_widget.py
flying-sheep: yeah, i think we should keep catching all exceptions and only in case of a ValueError (or probably a custom subclass we define) we swallow the error (as this one means that we got non-equation LaTeX)

ccordoba12: > Isn't it a good thing to log failure to render the latex?

Not for Spyder and Canopy because those failures are shown in the terminal where those applications were started. If they started graphically, the errors won't be seen by their users.

Besides, (as I said in ipython/ipython#9203) possible errors are:

1. Matplotlib is unable to handle Latex, something not easily discernible from the traceback.
2. Latex has syntax errors, something that also generates one of those long and almost unreadable report errors.

So this simplifies things and handles all cases I mentioned at the beginning correctly.

ccordoba12: > or probably a custom subclass we define

Yes, I was thinking to create a simple `LatexError` class.

qtconsole/rich_jupyter_widget.py
ccordoba12: The only that `latex_to_png` can handle is Latex enclosed in dollar signs, so there's no need to check for more at the moment.

qtconsole/rich_jupyter_widget.py
flying-sheep: good points. it’s the latex-repr-writers’ responsibility to emit correct code anyway and if they don’t, they’ll have to debug it.

qtconsole/rich_jupyter_widget.py
flying-sheep: [`latex_to_png_dvipng`](https://github.com/ipython/ipython/blob/9dbb9b2dea18e4ea9421c7a0ea226ce202430d6c/IPython/lib/latextools.py#L114) (called if [`backend == 'dvipng'`](https://github.com/ipython/ipython/blob/9dbb9b2dea18e4ea9421c7a0ea226ce202430d6c/IPython/lib/latextools.py#L87)) handles more right?

asmeurer:
Regarding the last checkmark, it's worth noting that matplotlib's math renderer produces output that's considerably worse than latex. Example:

**matplotlib**

matplotlib_math

**latex**

latex_math

For SymPy we prefer latex and only fallback to matplotlib if it isn't installed.

minrk:
@asmeurer thanks, following sympy's lead seems sensible here.

ccordoba12:
Thanks @asmeurer for the input. I changed the task accordingly.

flying-sheep:
OK, then the only thing left is the other mathmode triggers:

inline:

- `\(...\)`
- `$...$`
- environments (`\begin{name}...\end{name}`):
	- \begin{math}...\end{math}

display:

- \[...\]`
- $$...$$`
- environments (`\begin{name}...\end{name}`):
	- `displaymath`
	- `equation`/`equation*`
	- `eqnarray`/`eqnarray*`
	- `multline`/`multline*`
	- `gather`/`gather*`
	- `align`/`align*`
	- `flalign`/`flalign*`
	- `alignat`/`alignat*`

ccordoba12:
@flying-sheep, I won't implement all that, for two reasons:

1. Checking for dollar signs is enough for now, i.e. to have Sympy, IRKernel and Pandas work as expected.
2. I don't have time for it.

Maybe you could do it after this PR is merged :-)

ccordoba12:
Or you could send a PR against my branch if you want it as part of this.

flying-sheep:
no separately is OK

ccordoba12:
Against my branch? Please be my guest :-)

In any case, thanks for helping me out.

flying-sheep:
OK. here it is. is this sufficient or should i do the PR?

```python
basic_envs = 'math displaymath'.split()
starable_envs = 'equation eqnarray multline gather align flalign alignat'.split()
star_envs = [env + '*' for env in starable_envs]
envs = basic_envs + starable_envs + star_envs

env_syntax = [r'\begin{{{0}}} \end{{{0}}}'.format(env).split() for env in envs]

math_syntax = [
    (r'\[', r'\]'), (r'\(', r'\)'),
    ('$$', '$$'), ('$', '$'),
]

def is_latex_math(latex):
    for start, end in math_syntax + env_syntax:
        if latex.startswith(start) and latex.endswith(end):
            return True
        return False
```

ccordoba12:
I think it's enough, thanks a lot!

flying-sheep:
NP!

i edited it a bit since first commenting, but the version at the time of writing should be good

ccordoba12:
@minrk, I think this one is ready. Please review it and let me know what you think.

flying-sheep:
great, thanks!

Issue #96, opened 2016-02-06
the launcher on windows keeps a dos window open


stonebig:
Hi, 

When we double-click on "jupyter-qtconsole" on windows, there is a background black window that remains open.

![qt-console_black_window](https://cloud.githubusercontent.com/assets/4312421/12868797/7ffeebc8-cd0e-11e5-805e-a878b76ee2ca.GIF)

It was already present in QtConsole 3.2.0. Is there a small possible fix for that ?

stonebig:
hum,... I fail to reproduce with current master . closing

PR #97, opened 2016-02-06
avoid background DOS windows

stonebig:
resolve https://github.com/jupyter/qtconsole/issues/96 (on windows)


eric-wieser:
This causes the program to open in the background for me, which is a little annoyin.

Also worth noting that this only fixes `jupyter-qtconsole`, not `jupyter qtconsole`

PR #98, opened 2016-02-14
Remove %guiref and add instead an entry in our Help menu to show its contents

### LINKED ISSUE #78: The magic command %guiref not available in IPython 4, but is listed in QtConsole startup text
The line magic function %guiref is no longer available in IPython (version 4.0 and above I think, see the IPython [docs](http://ipython.readthedocs.org/en/stable/interactive/magics.html)), and should probably be removed from the text printed when starting Jupyter QtConsole, or replaced with a working command.

When I start Jupyter QtConsole (I'm on Windows 10 64 bit, using the latest Anaconda3) the following is printed

    Jupyter QtConsole 4.1.1
    Python 3.5.1 |Anaconda 2.4.1 (64-bit)| (default, Dec  7 2015, 15:00:12) [MSC v.1900 64 bit (AMD64)]
    Type "copyright", "credits" or "license" for more information.

    IPython 4.0.1 -- An enhanced Interactive Python.
    ?         -> Introduction and overview of IPython's features.
    %quickref -> Quick reference.
    help      -> Python's own help system.
    object?   -> Details about 'object', use 'object??' for extra details.
    %guiref   -> A brief reference about the graphical user interface.

But when I try %guiref I get an error

    In [1]: %guiref
    ERROR: Line magic function `%guiref` not found.

ccordoba12:
Fixes #78 

----

I'll create a corresponding pull request in the IPython repo to remove `gui_reference` from it.

### DIFF
--- a/qtconsole/jupyter_widget.py
+++ b/qtconsole/jupyter_widget.py
@@ -127,11 +127,10 @@ class JupyterWidget(IPythonWidget):
         else:
             self.set_default_style()
 
-        self._guiref_loaded = False
-
     #---------------------------------------------------------------------------
     # 'BaseFrontendMixin' abstract interface
     #---------------------------------------------------------------------------
+
     def _handle_complete_reply(self, rep):
         """Reimplemented to support Jupyter's improved completion machinery.
         """
@@ -270,11 +269,6 @@ class JupyterWidget(IPythonWidget):
         """Handle kernel info replies."""
         content = rep['content']
 
-        if not self._guiref_loaded:
-            if content['implementation'] == 'ipython':
-                self._load_guiref_magic()
-            self._guiref_loaded = True
-
         self.kernel_banner = content.get('banner', '')
         if self._starting:
             # finish handling started channels
@@ -282,24 +276,14 @@ class JupyterWidget(IPythonWidget):
             super(JupyterWidget, self)._started_channels()
 
     def _started_channels(self):
-        """Reimplemented to make a history request and load %guiref."""
+        """Reimplemented to make a history request"""
         self._starting = True
-        # The reply will trigger %guiref load provided
-        # implementation=='ipython'
-        self.kernel_client.kernel_info()
         self.kernel_client.history(hist_access_type='tail', n=1000)
 
 
     #---------------------------------------------------------------------------
     # 'FrontendWidget' protected interface
     #---------------------------------------------------------------------------
-
-    def _load_guiref_magic(self):
-        code = """from qtconsole import usage as _usage
-get_ipython().register_magic_function(_usage.page_guiref, 'line', 'guiref')
-del _usage
-        """
-        self.execute(code, hidden=True)
 
     def _process_execute_error(self, msg):
         """Handle an execute_error message"""
--- a/qtconsole/jupyter_widget.py
+++ b/qtconsole/jupyter_widget.py
@@ -129,10 +129,13 @@ class JupyterWidget(IPythonWidget):
 
     #---------------------------------------------------------------------------
     # 'BaseFrontendMixin' abstract interface
+    #
+    # For JupyterWidget,  override FrontendWidget methods which implement the
+    # BaseFrontend Mixin abstract interface
     #---------------------------------------------------------------------------
 
     def _handle_complete_reply(self, rep):
-        """Reimplemented to support Jupyter's improved completion machinery.
+        """Support Jupyter's improved completion machinery.
         """
         self.log.debug("complete: %s", rep.get('content', ''))
         cursor = self._get_cursor()
@@ -165,7 +168,7 @@ class JupyterWidget(IPythonWidget):
             self._complete_with_items(cursor, matches)
 
     def _handle_execute_reply(self, msg):
-        """ Reimplemented to support prompt requests.
+        """Support prompt requests.
         """
         msg_id = msg['parent_header'].get('msg_id')
         info = self._request_info['execute'].get(msg_id)
@@ -181,7 +184,7 @@ class JupyterWidget(IPythonWidget):
             super(JupyterWidget, self)._handle_execute_reply(msg)
 
     def _handle_history_reply(self, msg):
-        """ Implemented to handle history tail replies, which are only supported
+        """ Handle history tail replies, which are only supported
             by Jupyter kernels.
         """
         content = msg['content']
@@ -276,7 +279,7 @@ class JupyterWidget(IPythonWidget):
             super(JupyterWidget, self)._started_channels()
 
     def _started_channels(self):
-        """Reimplemented to make a history request"""
+        """Make a history request"""
         self._starting = True
         self.kernel_client.history(hist_access_type='tail', n=1000)
 
--- a/qtconsole/jupyter_widget.py
+++ b/qtconsole/jupyter_widget.py
@@ -269,7 +269,12 @@ class JupyterWidget(IPythonWidget):
     def _handle_kernel_info_reply(self, rep):
         """Handle kernel info replies."""
         content = rep['content']
-        
+
+        if not self._guiref_loaded:
+            if content['implementation'] == 'ipython':
+                self._load_guiref_magic()
+            self._guiref_loaded = True
+
         self.kernel_banner = content.get('banner', '')
         if self._starting:
             # finish handling started channels
@@ -279,15 +284,22 @@ class JupyterWidget(IPythonWidget):
     def _started_channels(self):
         """Reimplemented to make a history request and load %guiref."""
         self._starting = True
-        # The reply will trigger %guiref load provided language=='python'
+        # The reply will trigger %guiref load provided
+        # implementation=='ipython'
         self.kernel_client.kernel_info()
-
         self.kernel_client.history(hist_access_type='tail', n=1000)
-    
+
 
     #---------------------------------------------------------------------------
     # 'FrontendWidget' protected interface
     #---------------------------------------------------------------------------
+
+    def _load_guiref_magic(self):
+        code = """from qtconsole import usage as _usage
+get_ipython().register_magic_function(_usage.page_guiref, 'line', 'guiref')
+del _usage
+        """
+        self.execute(code, hidden=True)
 
     def _process_execute_error(self, msg):
         """Handle an execute_error message"""
--- a/qtconsole/usage.py
+++ b/qtconsole/usage.py
@@ -1,0 +1,221 @@
+"""
+Usage information for QtConsole
+"""
+
+# Copyright (c) Jupyter Development Team.
+# Distributed under the terms of the Modified BSD License.
+
+
+def page_guiref(arg_s=None):
+    """Show a basic reference about QtConsole."""
+    from IPython.core import page
+    page.page(gui_reference)
+
+
+gui_reference = """\
+=====================
+The Jupyter QtConsole
+=====================
+
+This console is designed to emulate the look, feel and workflow of a terminal
+environment, while adding a number of enhancements that are simply not possible
+in a real terminal, such as inline syntax highlighting, true multiline editing,
+inline graphics and much more.
+
+This quick reference document contains the basic information you'll need to
+know to make the most efficient use of it.  For the various command line
+options available at startup, type ``jupyter qtconsole --help`` at the command
+line.
+
+
+Multiline editing
+=================
+
+The graphical console is capable of true multiline editing, but it also tries
+to behave intuitively like a terminal when possible.  If you are used to
+IPython's old terminal behavior, you should find the transition painless, and
+once you learn a few basic keybindings it will be a much more efficient
+environment.
+
+For single expressions or indented blocks, the console behaves almost like the
+terminal IPython: single expressions are immediately evaluated, and indented
+blocks are evaluated once a single blank line is entered::
+
+    In [1]: print "Hello Jupyter!"  # Enter was pressed at the end of the line
+    Hello Jupyter!
+
+    In [2]: for i in range(10):
+       ...:     print i,
+       ...:
+    0 1 2 3 4 5 6 7 8 9
+
+If you want to enter more than one expression in a single input block
+(something not possible in the terminal), you can use ``Control-Enter`` at the
+end of your first line instead of ``Enter``.  At that point the console goes
+into 'cell mode' and even if your inputs are not indented, it will continue
+accepting arbitrarily many lines until either you enter an extra blank line or
+you hit ``Shift-Enter`` (the key binding that forces execution).  When a
+multiline cell is entered, the console analyzes it and executes its code producing
+an ``Out[n]`` prompt only for the last expression in it, while the rest of the
+cell is executed as if it was a script.  An example should clarify this::
+
+    In [3]: x=1  # Hit C-Enter here
+       ...: y=2  # from now on, regular Enter is sufficient
+       ...: z=3
+       ...: x**2  # This does *not* produce an Out[] value
+       ...: x+y+z  # Only the last expression does
+       ...:
+    Out[3]: 6
+
+The behavior where an extra blank line forces execution is only active if you
+are actually typing at the keyboard each line, and is meant to make it mimic
+the IPython terminal behavior.  If you paste a long chunk of input (for example
+a long script copied form an editor or web browser), it can contain arbitrarily
+many intermediate blank lines and they won't cause any problems.  As always,
+you can then make it execute by appending a blank line *at the end* or hitting
+``Shift-Enter`` anywhere within the cell.
+
+With the up arrow key, you can retrieve previous blocks of input that contain
+multiple lines.  You can move inside of a multiline cell like you would in any
+text editor.  When you want it executed, the simplest thing to do is to hit the
+force execution key, ``Shift-Enter`` (though you can also navigate to the end
+and append a blank line by using ``Enter`` twice).
+
+If you've edited a multiline cell and accidentally navigate out of it with the
+up or down arrow keys, the console will clear the cell and replace it with the
+contents of the one above or below that you navigated to.  If this was an
+accident and you want to retrieve the cell you were editing, use the Undo
+keybinding, ``Control-z``.
+
+
+Key bindings
+============
+
+The Jupyter QtConsole supports most of the basic Emacs line-oriented keybindings,
+in addition to some of its own.
+
+The keybinding prefixes mean:
+
+- ``C``: Control
+- ``S``: Shift
+- ``M``: Meta (typically the Alt key)
+
+The keybindings themselves are:
+
+- ``Enter``: insert new line (may cause execution, see above).
+- ``C-Enter``: *force* new line, *never* causes execution.
+- ``S-Enter``: *force* execution regardless of where cursor is, no newline added.
+- ``Up``: step backwards through the history.
+- ``Down``: step forwards through the history.
+- ``S-Up``: search backwards through the history (like ``C-r`` in bash).
+- ``S-Down``: search forwards through the history.
+- ``C-c``: copy highlighted text to clipboard (prompts are automatically stripped).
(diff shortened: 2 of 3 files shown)


takluyver:
This PR is still registering `%guiref` as a magic. I was initially thinking of a more dramatic change, where we don't involve the kernel in showing it at all, and just make an item in the help menu that brings it up. I know some of the text refers specifically to IPython, though - do you think it's feasible to rewrite these bits so the references to IPython are more examples than prescription? Or is it too tightly wedded to IPython to make any sense for other kernels?

ccordoba12:
> This PR is still registering %guiref as a magic

Yes, but only for `ipykernel`. As far as I know, `ipykernel` is the only kernel that shows `%guiref` as part of its banner. i tried to see if I could remove that line but I didn't understand how to do it.

However, I think it's ok to define `%guiref` when using `ipykernel` because

1. `ipykernel` is able to handle magics
2. People already know about the existence of `%guiref`.

> and just make an item in the help menu that brings it up

I also did that as part of this pull request.

> do you think it's feasible to rewrite these bits so the references to IPython are more examples than prescription?

I removed most of the references to IPython in `gui_reference`, and updated some imports too (like `IPython.core.display` to `IPython.display`). Please review it to see if things can be further improved :-)

takluyver:
I think I'd prefer to remove the `%guiref` magic entirely:

- It's only applicable to one frontend, so we shouldn't be showing it in the general banner when starting an IPython kernel. The banner should also show up in the terminal frontend (though it doesn't right now: jupyter/jupyter_console#59), and it's supposed to be independent of the frontend.
- It only works if the Qt console is installed in the same Python that the kernel is running in, which is by no means guaranteed.
- It's one more special case we can get rid of for our own kernel.

The banner that mentions `%guiref` is defined in `IPython.core.usage` and used in `ipykernel.zmqshell`. I'm happy to clear up both of those modules if you want to remove the magic registration in this PR.

ccordoba12:
> It only works if the Qt console is installed in the same Python that the kernel is running in

You're totally right. It makes no sense to maintain `%guiref` because it's not entirely dependent on the kernel.

So please remove its references from `ipykernel` and `IPython.core.usage`, then I'll remove it from here.

----

*Note*: The thing that confused me is that several things about `%guiref` are still present in qtconsole, so I assumed you removed it in a rebase or something. This time I'll make sure to remove all supporting code about it :-)

takluyver:
Removed in ipython/ipykernel#101 and ipython/ipython#9231

ccordoba12:
Ok, I think this is ready, unless you have some comments about my changes to `gui_reference`.

ccordoba12:
`%guiref` also appears in this [figure](https://github.com/jupyter/qtconsole/blob/master/docs/source/figs/jn.xhtml), so it should be updated accordingly. Maybe @willingc could help us with that?

qtconsole/usage.py
```diff
@@ -0,0 +1,221 @@
...
+
+
+gui_reference = """\
+=====================
+The Jupyter QtConsole
+=====================
+
+This console is designed to emulate the look, feel and workflow of a terminal
+environment, while adding a number of enhancements that are simply not possible
+in a real terminal, such as inline syntax highlighting, true multiline editing,
```
takluyver: We might want to change the phrasing, because the first two of these (syntax highlighting and multiline editing) are coming to IPython for version 5 (see ipython/ipython#9118). And even inline graphics are [possible](https://github.com/saitoha/PySixel), albeit not supported by most terminal emulators.

qtconsole/usage.py
```diff
@@ -0,0 +1,221 @@
...
+
+For single expressions or indented blocks, the console behaves almost like the
+terminal IPython: single expressions are immediately evaluated, and indented
+blocks are evaluated once a single blank line is entered::
+
+    In [1]: print "Hello Jupyter!"  # Enter was pressed at the end of the line
+    Hello Jupyter!
+
+    In [2]: for i in range(10):
+       ...:     print i,
```
takluyver: Let's make our examples Python 3 compliant, to set a good example. And so they don't look like they haven't been updated in years... ;-)

qtconsole/usage.py
```diff
@@ -0,0 +1,221 @@
...
+(something not possible in the terminal), you can use ``Control-Enter`` at the
+end of your first line instead of ``Enter``.  At that point the console goes
+into 'cell mode' and even if your inputs are not indented, it will continue
+accepting arbitrarily many lines until either you enter an extra blank line or
+you hit ``Shift-Enter`` (the key binding that forces execution).  When a
+multiline cell is entered, the console analyzes it and executes its code producing
+an ``Out[n]`` prompt only for the last expression in it, while the rest of the
+cell is executed as if it was a script.  An example should clarify this::
+
+    In [3]: x=1  # Hit C-Enter here
```
takluyver: We should spell out 'Ctrl-', rather than relying on the emacs 'C-' convention.

qtconsole/usage.py
```diff
@@ -0,0 +1,221 @@
...
+============
+
+The Jupyter QtConsole supports most of the basic Emacs line-oriented keybindings,
+in addition to some of its own.
+
+The keybinding prefixes mean:
+
+- ``C``: Control
+- ``S``: Shift
+- ``M``: Meta (typically the Alt key)
```
takluyver: Again, my preference is for explicit Ctrl/Shift/Alt in the list below, rather than defining abbreviations.

qtconsole/usage.py
```diff
@@ -0,0 +1,221 @@
...
+
+The Jupyter QtConsole will show long blocks of text from many sources using a
+builtin pager. You can control where this pager appears with the ``--paging``
+command-line flag:
+
+- ``inside`` [default]: the pager is overlaid on top of the main terminal. You
+  must quit the pager to get back to the terminal (similar to how a pager such
+  as ``less`` or ``more`` works).
+
+- ``vsplit``: the console is made double-tall, and the pager appears on the
```
takluyver: Maybe it's just me, but 'double-tall' and 'double-wide' don't sound quite right. I'd say 'double height' and 'double width'.

qtconsole/usage.py
```diff
@@ -0,0 +1,221 @@
...
+  bottom area when needed.  You can view its contents while using the terminal.
+
+- ``hsplit``: the console is made double-wide, and the pager appears on the
+  right area when needed.  You can view its contents while using the terminal.
+
+- ``none``: the console never pages output.
```
takluyver: This isn't very clear: it presumably means that long text that would be paged is displayed like output within the flow of the console, but it could suggest that that output is just discarded.

qtconsole/usage.py
```diff
@@ -0,0 +1,221 @@
...
+- Control-o: goes from one to another always.
+- Mouse: click on either.
+
+In all cases, the ``q`` or ``Escape`` keys quit the pager (when used with the
+focus on the pager area).
+
+Running subprocesses
+====================
+
+The Jupyter QtConsole uses the ``pexpect`` module to run subprocesses
```
takluyver: It's the IPython kernel that uses Pexpect. But this technical detail is irrelevant to the main point: subprocesses run from the kernel cannot interactively prompt the user for input.

There's a bit of a tricky distinction for new programmers - If you run `foo.py` with something like `%run` or `execfile()`, it is running in the kernel process and *can* get simple input from the user (though full terminal interaction like less or vim is still impossible). But if you run it with `!python`, it's a separate process and cannot prompt the user for any input. I don't know how best to convey this.

qtconsole/usage.py
```diff
@@ -0,0 +1,221 @@
...
+when you type ``!command``.  This has a number of advantages (true asynchronous
+output from subprocesses as well as very robust termination of rogue
+subprocesses with ``Control-C``), as well as some limitations.  The main
+limitation is that you can *not* interact back with the subprocess, so anything
+that invokes a pager or expects you to type input into it will block and hang
+(you can kill it with ``Control-C``).
+
+We have provided as magics ``%less`` to page files (aliased to ``%more``),
```
takluyver: This should make it clear that these magics are an IPython kernel feature, not something the Qt console itself provides.

qtconsole/usage.py
```diff
@@ -0,0 +1,221 @@
...
+``%clear`` to clear the terminal, and ``%man`` on Linux/OSX.  These cover the
+most common commands you'd want to call in your subshell and that would cause
+problems if invoked via ``!cmd``, but you need to be aware of this limitation.
+
+Display
+=======
+
+The Jupyter QtConsole can now display objects in a variety of formats, including
+PNG and SVG. This is accomplished using the display functions in
```
takluyver: Again, this should make it clear that it's an example of how to do it from IPython, and that it may be different for other kernels.

Also, I think we can lose the word 'now' in the line above.

qtconsole/usage.py
```diff
@@ -0,0 +1,221 @@
...
+    In [8]: png_formatter.for_type(Foo, foo_to_png)
+
+For further details, see ``IPython.core.formatters``.
+
+Inline matplotlib graphics
+==========================
+
+The Jupyter QtConsole is capable of displaying matplotlib figures inline, in SVG
+or PNG format.  If started with the ``matplotlib=inline``, then all figures are
+rendered inline automatically (PNG by default).  If started with ``--matplotlib``
```
takluyver: This flag doesn't even work any more. I think we should just remove this section - the integration with mpl is really an IPython feature, not a Qt console feature.

takluyver:
I know most of the things I'm pointing out in the text are not things you changed, but that text is long overdue for an update, and this seems like a good opportunity to go through it. Bad luck!

@willingc , could you also do a pass on this to help make it clearer? Thanks.

qtconsole/jupyter_widget.py
```diff
@@ -277,13 +276,10 @@ def _handle_kernel_info_reply(self, rep):
             super(JupyterWidget, self)._started_channels()
 
     def _started_channels(self):
-        """Reimplemented to make a history request and load %guiref."""
+        """Reimplemented to make a history request"""
```
willingc: Perhaps a semantic quibble on my part but "reimplemented" feels not quite accurate to me. BaseFrontendWidget is abstract; FrontEndWidget implements the method from the abstract BaseFrontEndWidget; JupyterWidget subclasses from FrontEndWidget.

The docstrings for each method in this section are not consistent. "Override method" or "Redefine method" feel more accurate to me. Now that I'm thinking about it, it may just be best to add a comment here:
```
 #---------------------------------------------------------------------------
 # 'BaseFrontendMixin' abstract interface
 #
 # For JupyterWidget,  override FrontendWidget methods which implement the
 # BaseFrontend Mixin abstract interface
 #---------------------------------------------------------------------------
 ```

File: qtconsole/usage.py [hunk 12]
```diff
@@ -0,0 +1,221 @@
+"""
+Usage information for QtConsole
+"""
+
+# Copyright (c) Jupyter Development Team.
+# Distributed under the terms of the Modified BSD License.
+
+
+def page_guiref(arg_s=None):
+    """Show a basic reference about QtConsole."""
```
willingc: """Show reference text about QtConsole usage."""

qtconsole/usage.py
willingc: The console is designed to emulate the look, feel and workflow of a terminal environment. Beyond this basic design, the console may also implement functionality not currently found in most terminal emulators. Some examples of these console enhancements are inline syntax highlighting, multiline editing, inline graphics, and others.

qtconsole/usage.py
```diff
@@ -0,0 +1,221 @@
...
+=================
+
+The graphical console is capable of true multiline editing, but it also tries
+to behave intuitively like a terminal when possible.  If you are used to
+IPython's old terminal behavior, you should find the transition painless, and
+once you learn a few basic keybindings it will be a much more efficient
+environment.
+
+For single expressions or indented blocks, the console behaves almost like the
+terminal IPython: single expressions are immediately evaluated, and indented
```
willingc: IPython terminal

qtconsole/usage.py
```diff
@@ -0,0 +1,221 @@
...
+options available at startup, type ``jupyter qtconsole --help`` at the command
+line.
+
+
+Multiline editing
+=================
+
+The graphical console is capable of true multiline editing, but it also tries
+to behave intuitively like a terminal when possible.  If you are used to
+IPython's old terminal behavior, you should find the transition painless, and
```
willingc: ... painless. If you learn to use a few basic keybindings, the console provides even greater efficiency.

qtconsole/usage.py
```diff
@@ -0,0 +1,221 @@
...
+once you learn a few basic keybindings it will be a much more efficient
+environment.
+
+For single expressions or indented blocks, the console behaves almost like the
+terminal IPython: single expressions are immediately evaluated, and indented
+blocks are evaluated once a single blank line is entered::
+
+    In [1]: print "Hello Jupyter!"  # Enter was pressed at the end of the line
```
willingc: print("Hello Jupyter!")

qtconsole/usage.py
```diff
@@ -0,0 +1,221 @@
...
+    Hello Jupyter!
+
+    In [2]: for i in range(10):
```
willingc: for num in range(10):
    print(num)

qtconsole/usage.py
```diff
@@ -0,0 +1,221 @@
...
+       ...:     print i,
+       ...:
+    0 1 2 3 4 5 6 7 8 9
+
+If you want to enter more than one expression in a single input block
+(something not possible in the terminal), you can use ``Control-Enter`` at the
+end of your first line instead of ``Enter``.  At that point the console goes
```
willingc: Huh, didn't know that :-)

qtconsole/usage.py
```diff
@@ -0,0 +1,221 @@
...
+into 'cell mode' and even if your inputs are not indented, it will continue
+accepting arbitrarily many lines until either you enter an extra blank line or
```
willingc: accepting ~~arbitrarily many~~ lines

qtconsole/usage.py
willingc: +1 to @takluyver's suggestion

qtconsole/usage.py
```diff
@@ -0,0 +1,221 @@
...
+you can then make it execute by appending a blank line *at the end* or hitting
+``Shift-Enter`` anywhere within the cell.
+
+With the up arrow key, you can retrieve previous blocks of input that contain
+multiple lines.  You can move inside of a multiline cell like you would in any
+text editor.  When you want it executed, the simplest thing to do is to hit the
+force execution key, ``Shift-Enter`` (though you can also navigate to the end
+and append a blank line by using ``Enter`` twice).
+
+If you've edited a multiline cell and accidentally navigate out of it with the
```
willingc: Is this what is happening:

If you are editing a multiline cell and accidentally navigate out of it using the up or down arrow keys, the console clears the cell and replaces it with the contents of the cell which the up or down arrow key stopped on. If you wish to undo this action, perhaps because of an accidental keypress, use the Undo keybinding, ``Control-z``, to restore the original cell.

qtconsole/usage.py
willingc: +1

qtconsole/usage.py
```diff
@@ -0,0 +1,221 @@
...
+- ``C-.``: force a kernel restart (a confirmation dialog appears).
+- ``C-+``: increase font size.
+- ``C--``: decrease font size.
+- ``C-M-Space``: toggle full screen. (Command-Control-Space on Mac OS X)
+
+The pager
+=========
+
+The Jupyter QtConsole will show long blocks of text from many sources using a
+builtin pager. You can control where this pager appears with the ``--paging``
```
willingc: built-in

qtconsole/usage.py
```diff
@@ -0,0 +1,221 @@
...
+command-line flag:
+
+- ``inside`` [default]: the pager is overlaid on top of the main terminal. You
+  must quit the pager to get back to the terminal (similar to how a pager such
```
willingc: ...similar to how ``less`` or ``more`` pagers behave.)

qtconsole/usage.py
willingc: Just to clarify, is vsplit defined correctly as is? @takluyver, @ccordoba12 or should it be replaced by hsplit? They seem backwards to me.

Suggestion:
The console is split vertically. The pager appears in the bottom area when needed. You can view the pager's contents while using the terminal.

Similar text for hsplit.

qtconsole/usage.py
willingc: Perhaps ``none``: the console displays output without paging.

qtconsole/usage.py
willingc: The console can use magics provided by the IPython kernel. These magics include:  ``%less`` ...

qtconsole/usage.py
willingc: For example, if using the IPython kernel, there are functions available for object display:

qtconsole/usage.py
willingc: I agree lets remove this entire section of "Inline matplotlib graphics"

willingc:
@ccordoba12 @takluyver I've gone through and added my comments. Depending on the time window for 4.2 release, it might be good to merge after these changes are made. After that I would be happy to iterate any further changes as needed.

qtconsole/usage.py
takluyver: Ah, I remember this debate. There is approximately no consensus on which way round the terms 'horizontal split' and 'vertical split' should be applied. To my mind, it does make sense that vsplit puts the pager below, but I can see why other people disagree.

At one point I think we planned to rename them to 'bottom' and 'right', which would avoid this ambiguity. But I guess that never quite happened.

qtconsole/usage.py
ccordoba12: I also don't know how to address this.

ccordoba12:
Thanks for your review. I addressed all comments, except for one from @takluyver about the use of `subprocess`.

qtconsole/usage.py
takluyver: I think I'd just remove the first couple of sentences, and start with: "When running a subprocess from the kernel, you can not interact with it as if it was running in a terminal..."

willingc:
@ccordoba12 Looks great! I noticed @takluyver had a wording suggestion. Other than that, this is ready to merge.

qtconsole/usage.py
Carreau: OK, I think it's great. If you want to just update this bit and merge (feel free to self merge). Otherwise, I'll merge next time I come by this thread.

ccordoba12:
@minrk or @takluyver, I think everything is ready now for you to release 4.2 :-)

willingc:
@ccordoba12 :cake: :cookie:

minrk:
@ccordoba12 released 4.2! Thanks for your help.

ccordoba12:
Thanks a lot Min!!

temok-mx:
Dear All,

Wishes for the QtConsole:
1)   Enable searches like in Emacs (`control-s`) or like in the `less` command (slash '/').
This would be useful in situations like when browsing the QtConsole Help.

2)   Enable the display of the QtConsole Help from a command in the QtConsole (in addition or in replacement of using the mouse to click on the Help menu).

Cheers,
Temok

ccordoba12:
Regarding your points @temok-mx:

1) This is implemented in Spyder, i.e. if you press Ctrl+F you get a search dialog to easily look for something in Qtconsole content.

2) Previously Qtconsole had a command called `%guiref` to do precisely this, but it was removed by this PR. The reasoning behind this decision is described in https://github.com/jupyter/qtconsole/pull/98#issuecomment-184202984

3) That's a good idea, but someone will have to implement it. You could open a new issue about it :-)

temok-mx:
Thanks Carlos,

I just posted an issue for point 3):  #122 
Saludos,
Temok

PR #110, opened 2016-03-14
Home key event updated.

nisseg:
Move cursor to beginning of text instead of beginning of line
in the case line starts with leading spaces. This is the same
behaviour used by common text editors/IDEs like SublimeText
and PyCharm.

### DIFF
--- a/qtconsole/console_widget.py
+++ b/qtconsole/console_widget.py
@@ -1371,10 +1371,13 @@ class ConsoleWidget(MetaQObjectHasTraits('NewBase', (LoggingConfigurable, QtGui.
                     start_pos = cursor.position()
                     start_pos += len(self._continuation_prompt)
                     cursor.setPosition(position)
+
                 c = self._get_cursor()
                 spaces = self._get_leading_spaces()
-                if c.position() > start_pos + spaces:
-                    start_pos += spaces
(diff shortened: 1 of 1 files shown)


takluyver:
The behaviour I normally see is that the first press goes to the start of the text (i.e. after any leading spaces), and then subsequent presses alternate between there and the start of the line (before leading spaces).

nisseg:
takluyver, I see I left that part out in the description. I.e. the bit when alternating from start of text to start of line. Actually, I think it is implemented according to the behavior you describe.

takluyver:
I think it will go to start-of-line on the second press, but I don't see what would make it go back to start-of-text on the third (fifth, seventh, etc.) press?

nisseg:
Ah ok, I see what you mean. I'll close and fix. Thanks for the comment.

takluyver:
You don't need to close a PR to make further changes - you can push another commit to the same branch, and it will update automatically.

nisseg:
Alright, I just pushed a new commit. The cursor will now keep alternating between start of line and start of text. Much better :+1:

takluyver:
Thanks! The code looks OK to me, but I'll let @ccordoba12 have a look.

minrk:
:+1:, I'll let @ccordoba12 merge.

PR #111, opened 2016-03-14
Enter, Return key event modified

nisseg:
The Enter/Return key event is updated to also indent new lines
within an existing text block. Currently only lines added at the
bottom of the console are indented.

### DIFF
--- a/qtconsole/console_widget.py
+++ b/qtconsole/console_widget.py
@@ -1166,10 +1166,12 @@ class ConsoleWidget(MetaQObjectHasTraits('NewBase', (LoggingConfigurable, QtGui.
                         self.execute(interactive = not shift_down)
                     else:
                         # Do this inside an edit block for clean undo/redo.
+                        indent = self._get_next_row_indent()
                         cursor.beginEditBlock()
                         cursor.setPosition(position)
                         cursor.insertText('\n')
                         self._insert_continuation_prompt(cursor)
+                        cursor.insertText(' ' * indent)
                         cursor.endEditBlock()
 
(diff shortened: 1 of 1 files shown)


qtconsole/console_widget.py
```diff
@@ -1660,6 +1662,24 @@ def _get_leading_spaces(self):
...
+        prompt = len(self._continuation_prompt)
+        cur_pos = cur.columnNumber() - prompt
+        cur.select(QtGui.QTextCursor.LineUnderCursor)
+        text = cur.selectedText()[prompt:]
+        leading_spaces = len(text) - len(text.lstrip())
+        if cur_pos < leading_spaces:
+            indent = cur_pos
+        else:
+            indent = leading_spaces
+            if 0 < len(text) <= cur_pos and text.rstrip()[-1] == ':':
```
minrk: This shouldn't be making any Python-specific syntax assumptions.

nisseg:
Sorry, for sneaking that in. I removed those Python syntax specific lines. It would however be a little bit nicer typing in the console having that indent. Are you reluctant to include any language specific syntaxing into this class or would it be OK with something like: `if lang == python:    do this`? Are there any other alternatives?

jdfreder:
There's a notion of completeness that kernels can and do implement for this type of check.  http://ipython.org/ipython-doc/dev/development/messaging.html#code-completeness  Because kernels are responsible for implementing it and it's a part of the spec, it's language agnostic.

nisseg:
Thanks for the reference, that might have set me on track :+1:

minrk:
Thanks!

PR #112, opened 2016-03-18
Updated the conda build for ReadTheDocs

willingc:
Updates the conda `environment.yml` and `conf.py` files to build cleanly with conda and Python 3 on RTD. Also removed the commented numpydoc lines since they are not necessary when sphinx is pinned to 1.3.6 or greater. This should resolve the currently failing builds on RTD.


takluyver:
Let's see what happens.

takluyver:
Thanks @willingc, it's building again. However, the [config options doc](http://qtconsole.readthedocs.org/en/latest/config_options.html) is not showing any of the options for frontend classes. I guess that we need to install pyqt in the conda environment so that it can import the files and discover those options.

PR #113, opened 2016-03-18
Front end config options in docs

willingc:


docs/environment.yml
```diff
@@ -1,10 +1,11 @@
 name: qtconsole_docs
 channels:
+  - astrofrog
```
takluyver: I don't think we need an extra channel here - pyqt is in the standard conda channel.

willingc: Thanks @takluyver. Just a clarification on the front end config options: were there items missing or detail on the items (i.e. options) missing or both?

takluyver: Oh, they're there now. Maybe I had an old copy cached or something. I was seeing only the `Session` options in the doc - everything else was completely missing. I'm not really sure how it's working without pyqt installed, though :confused: .

willingc: Oh cool. It was pulling pyqt from the astrofrog channel. I've gone back to pulling from anaconda in a future commit :smile:

takluyver:
:cookie: :)

willingc:
Thanks for the reviews @takluyver :cookie:

willingc:
@takluyver One last thing on this... we need to wipe and rebuild stable on RTD so it uses the current conda environment.yml. I don't have administrative access on this repo on RTD so I can't do it. Happy to do it if you bump my privileges. Thanks!

takluyver:
Stable points to the latest release tag, so I don't think it will get the changes until we make a new release with these changes. But maybe you know something more?

willingc:
Actually stable is currently pointing to 4.1.1 (unless my cache is messed up). I believe we would like stable to point to 4.2.0, and latest to head of master (4.3.0dev).

takluyver:
Stable on the RTD admin panel is pointing to the 4.2.0, but I guess it has failed to build, so it's still showing the docs from 4.1.1. What's your RTD username? I'm happy to add you to the maintainers.

willingc:
willingc is the username. For future reference, if you go into the Versions tab and select Wipe next to stable, it should begin rebuilding stable with the latest release tag. Though sometimes it is finicky.
screen shot 2016-03-19 at 5 00 23 am

takluyver:
Right, but the latest release tag still has the same conda environment with which it failed to build before, so I'd expect it to fail again.

Added you as a maintainer

willingc:
@takluyver Thanks for the add. What is very odd is that the release tag SHA that RTD has 4.2.0 (fc3ed4d7) is different the actual release tag for 4.2.0 in the github repo (08bde28). Which leads me to think that RTD is not updating the tags properly (using git fetch --tags or git remote update). Weird.

takluyver:
The even stranger thing is that if I try to see what commit fc3ed4d7 is, it acts as if I asked for 08bde28:

```
git checkout fc3ed4d7
Previous HEAD position was 159cf30... Update tests
HEAD is now at 08bde28... release 4.2.0
```

I do not understand what's going on. Does git maybe calculate two different hashes?

willingc:
Hmm... I will do some research later this weekend. Strange though. On a productive note, I've made some more progress (though still a hack) on a [docs development dashboard](http://willingc.github.io/doc_helper/) that allows an easy glance at how our docs are building and a few other pieces of data. 
screen shot 2016-03-19 at 7 14 27 am

takluyver:
Neat!

PR #114, opened 2016-03-20
Make SuperQObject logic reusable with other Qt bases

### LINKED ISSUE #104: ConsoleWidget(): `RuntimeError: You can't initialize an object twice!` with traitlets master
```
from qtconsole.qt import QtGui
from qtconsole.console_widget import ConsoleWidget
QtGui.QApplication.instance() or QtGui.QApplication([])
cw = ConsoleWidget()
```
Results in 
```
  File "../../canopy-jupyter/scratch.py", line 4, in <module>
    cw = ConsoleWidget()
  File "/home/pankaj/work/enthought/jupyter/qtconsole/qtconsole/console_widget.py", line 214, in __init__
    LoggingConfigurable.__init__(self, **kw)
  File "/home/pankaj/work/enthought/jupyter/traitlets/traitlets/config/configurable.py", line 74, in __init__
    super(Configurable, self).__init__(**kwargs)
  File "/home/pankaj/work/enthought/jupyter/traitlets/traitlets/traitlets.py", line 946, in __init__
    super(HasTraits, self).__init__()
RuntimeError: You can't initialize an object twice!
```
with traitlets master. This is because of mixing old-style and `__init__` call and `super` `__init__` call.
Simple `git bisect` indicated the causal commit to be: https://github.com/ipython/traitlets/pull/174

A solution would be to do things similar to `SuperQObject` https://github.com/jupyter/qtconsole/blob/master/qtconsole/util.py#L48 , i.e. a `SuperQWidget`, and probably any other `Super` qobject subclass which needs to be used in multiple inheritance with HasTraits.

minrk:
And apply it to QWidget in ConsoleWidget.

Fixes double-initialization in ConsoleWidget with:

- PySide
- traitlets 4.2, which calls super in `__init__` now

closes #104

### DIFF
--- a/qtconsole/console_widget.py
+++ b/qtconsole/console_widget.py
@@ -208,6 +208,8 @@ class ConsoleWidget(MetaQObjectHasTraits('NewBase', (LoggingConfigurable, superQ
             The parent for this widget.
         """
         super(ConsoleWidget, self).__init__(**kw)
+        if parent:
+            self.setParent(parent)
 
         # While scrolling the pager on Mac OS X, it tears badly.  The
         # NativeGesture is platform and perhaps build-specific hence
--- a/qtconsole/console_widget.py
+++ b/qtconsole/console_widget.py
@@ -15,7 +15,7 @@ from qtconsole.qt import QtCore, QtGui
 
 from traitlets.config.configurable import LoggingConfigurable
 from qtconsole.rich_text import HtmlExporter
-from qtconsole.util import MetaQObjectHasTraits, get_font
+from qtconsole.util import MetaQObjectHasTraits, get_font, superQ
 from ipython_genutils.text import columnize
 from traitlets import Bool, Enum, Integer, Unicode
 from .ansi_code_processor import QtAnsiCodeProcessor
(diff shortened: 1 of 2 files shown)


ccordoba12:
I can confirm that this fixes the problem reported in #104 with PySide.

I'm not qualified to review this PR though :-)

qtconsole/console_widget.py
```diff
@@ -207,8 +207,7 @@ def __init__(self, parent=None, **kw):
         parent : QWidget, optional [default None]
             The parent for this widget.
         """
-        QtGui.QWidget.__init__(self, parent)
-        LoggingConfigurable.__init__(self, **kw)
+        super(ConsoleWidget, self).__init__(**kw)
```
pankajp: There's a problem here in that the `parent` `QWidget` is not set for self

minrk: Thanks. Can you tell me what problem that would cause? I've been using this for a while, and I can't see any misbehavior.

pankajp: Well i haven't really checked, but i assume it is violating the user's expectations that passing in a parent would set the parent widget. Quoting from the Qt docs for `QWidget::QWidget ( QWidget * parent = 0, Qt::WindowFlags f = 0 )`
> If parent is 0, the new widget becomes a window. If parent is another widget, this widget becomes a child window inside parent. The new widget is deleted when its parent is deleted

The following code should show a `ConsoleWidget` inside the `QWidget`, which fails if the parent is not set.
```
w = QtGui.QWidget()
cw = ConsoleWidget(w)
w.show()  # Should show a ConsoleWidget within the window
```

You can simply set the parent in here via a `self.setParent(parent)` call after the `__init__()` call. It actually is not a problem in practice because we all use the `ConsoleWidget` within a layout, which automatically manages the parenting of its child widgets.

minrk: Added a call to setParent. That's exactly what I needed, thanks!

pankajp: Thanks for the fix :+1:

minrk:
Thanks, @pankajp!

PR #125, opened 2016-05-05
BUG: Rename width/height traits to avoid QWidget override.

### LINKED ISSUE #124: ConsoleWidget width/height traits override Qt properties
https://github.com/jupyter/qtconsole/blob/master/qtconsole/console_widget.py#L135-L143

These override the `width()` and `height()` Qt properties that are required to use this as a `QWidget` in many places. Can we rename them to `console_width/console_height`? I don't know how often these traits are actually used downstream. We override them right back to work around the problem.

rkern:
Fixes #124

Not sure what should be done with the changelog to document this.

### DIFF
--- a/docs/source/changelog.rst
+++ b/docs/source/changelog.rst
@@ -3,6 +3,17 @@
 Changes in Jupyter Qt console
 =============================
 
+.. _4.3:
+
+4.3
+---
+
+`4.3 on GitHub <https://github.com/jupyter/qtconsole/milestones/4.3>`__
+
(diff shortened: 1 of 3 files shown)


takluyver:
Put a note in the changelog. Technically, since the change breaks API and config, it should be for 5.0, but clashing with Qt properties seems broken enough that I think it could go into 4.3.

Ping @ccordoba12 .

ccordoba12:
+1 from me :-)

takluyver:
OK, feel free to merge and decide what release it goes in :-)

rkern:
Thanks, all!

PR #133, opened 2016-08-21
Update jupyter_widget.py

dsblank:
Allow the msg content to not have a metadata component, which appears to be ignored anyway.

### DIFF
--- a/qtconsole/jupyter_widget.py
+++ b/qtconsole/jupyter_widget.py
@@ -259,7 +259,8 @@ class JupyterWidget(IPythonWidget):
         if self.include_output(msg):
             self.flush_clearoutput()
             data = msg['content']['data']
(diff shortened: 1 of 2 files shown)


takluyver:
[The spec](http://jupyter-client.readthedocs.io/en/latest/messaging.html#display-data) doesn't say that metadata is optional, and I don't think we intend it to be optional. Which kernel are you using when you run into this?

dsblank:
@takluyver Ok, thanks for the info. I had fixed an issue (https://github.com/Calysto/metakernel/issues/100) but I think I had misinterpreted that and removed metadata from too many locations. I think metakernel now doesn't have it where it doesn't belong (streams) and does have it where required (display_data). I'll close this.

Issue #135, opened 2016-08-30
Bug in case of input(int(239))


PolarNick239:
**Steps to reproduce:**
In qtconsole execute `input(239)`

**Expected:**
The same behaivour as in case of `input("239")`

**Found**:
Console does not respond to keyboard input anymore and traceback in terminal:
```
Traceback (most recent call last):
  File "/home/polarnick/libraries/python3.4/lib/python3.4/site-packages/qtconsole/base_frontend_mixin.py", line 163, in _dispatch
    handler(msg)
  File "/home/polarnick/libraries/python3.4/lib/python3.4/site-packages/qtconsole/frontend_widget.py", line 516, in _handle_input_request
    self._readline(msg['content']['prompt'], callback=callback, password=msg['content']['password'])
  File "/home/polarnick/libraries/python3.4/lib/python3.4/site-packages/qtconsole/console_widget.py", line 2010, in _readline
    self._show_prompt(prompt, newline=False)
  File "/home/polarnick/libraries/python3.4/lib/python3.4/site-packages/qtconsole/console_widget.py", line 2098, in _show_prompt
    self._append_plain_text(prompt)
  File "/home/polarnick/libraries/python3.4/lib/python3.4/site-packages/qtconsole/console_widget.py", line 924, in _append_plain_text
    self._append_custom(self._insert_plain_text, text, before_prompt)
  File "/home/polarnick/libraries/python3.4/lib/python3.4/site-packages/qtconsole/console_widget.py", line 894, in _append_custom
    result = insert(cursor, input, *args, **kwargs)
  File "/home/polarnick/libraries/python3.4/lib/python3.4/site-packages/qtconsole/console_widget.py", line 1761, in _insert_plain_text
    self._pending_insert_text, buffer_size)
  File "/home/polarnick/libraries/python3.4/lib/python3.4/site-packages/qtconsole/console_widget.py", line 1635, in _get_last_lines_from_list
    return_count=True)
  File "/home/polarnick/libraries/python3.4/lib/python3.4/site-packages/qtconsole/console_widget.py", line 1610, in _get_last_lines
    pos = len(text)
TypeError: object of type 'int' has no len()
Traceback (most recent call last):
  File "/home/polarnick/libraries/python3.4/lib/python3.4/site-packages/qtconsole/console_widget.py", line 1494, in _on_flush_pending_stream_timer
    self._flush_pending_stream()
[...]

PR #136, opened 2016-08-30
Ensure that msg['content']['prompt'] is string

PolarNick239:
Fix for https://github.com/jupyter/qtconsole/issues/135

### DIFF
--- a/qtconsole/frontend_widget.py
+++ b/qtconsole/frontend_widget.py
@@ -512,7 +512,7 @@ class FrontendWidget(HistoryConsoleWidget, BaseFrontendMixin):
         if self._reading:
             self.log.debug("Got second input request, assuming first was interrupted.")
             self._reading = False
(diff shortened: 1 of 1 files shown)


PolarNick239:
Ipython supports python2.7 only in 5.x versions, but [requirements.txt](https://github.com/jupyter/qtconsole/blob/master/requirements.txt) uses master version

May be it is possible to require `-e git+https://github.com/ipython/ipython.git@5.1.0#egg=ipython` exactly for python2.7?

minrk:
The protocol declares that this will be a string, so I think the bug is really on the IPython kernel side, where we should be doing the same casting that stdlib raw_input does. ipython/ipykernel#186 does this.

I've opened #137 to address the test dependencies problem. That requirements.txt should have gone away a long time ago.

PolarNick239:
Ok, thanks!

Issue #140, opened 2016-09-07
Exceptions not showing up when running a script with ipython magic %run


tleonhardt:
So if I use the IPython magic "%run" command to execute some basic code in a Python script, lets say our script contains only this code:
    1 + 'foo'

This code will should generate a TypeError Exception.  If I use %run to execute the script, then the proper exception appers in ipython at the command-line or in a Jupyter Notebook.  However, if I execute the exact same code in a Jupyter qtconsole, it doesn't show any exception.

I have tested and verified this bug using the latest version of ipython, ipykernel, jupyter_console and all of the other jupyter stuff from pip.  I have also verified it occurs on Windows, Mac, and Linux.  This also affects jupyter qtconsoles embedded in other applications like Spyder, PyCharm, or JupyterLab.

wmvanvliet:
The messaging protocol for errors changes slightly.

see #141

wmvanvliet:
fixed in #141

brauliobarahona:
Hi all, is this fixed in `Jupyter QtConsole 4.2.1, Python 3.5.2` ? 

When running a script from the Jupyter QtConsole with `run -i script_name.py` the scripts seems to execute fine, but it actually stops when it reaches a line looking for a non existent variable `YE`.

When doing the same in `Jupyter console 5.0.0, Python 3.5.2 ` I get the exception `NameError: name 'YE' is not defined` as expected.

My installation is on MacOs with Anaconda 4.3.1.

ccordoba12:
> is this fixed in Jupyter QtConsole 4.2.1, Python 3.5.2 ?

No, it was fixed in QtConsole 4.3. Please update.

PR #141, opened 2016-09-13
[MRG] Fix error messages when using %run

### LINKED ISSUE #156: Missing error messages when using PyQt5 event loop integration
I'm trying to have exceptions from a Qt application print to the Jupyter QtConsole and also have the application not block input to the QtConsole (event loop integration). This was working with PyQt4 and IPython 4.2, but with PyQt5 and IPython 5.1 I can either have exceptions printed and block input, or have no exceptions printed with event loop integration.

In Jupyter QtConsole, running the following and clicking the button doesn't print the exception. If `%gui qt5` is skipped, it does print the exception but the application blocks the console.
```
%gui qt5
run test.py
```

test.py:

```python
from PyQt5.QtWidgets import QWidget, QPushButton, QApplication
import sys

class MyWidget(QWidget):
    def __init__(self, parent=None):
        super(MyWidget, self).__init__(parent)
        self.pushButton = QPushButton('test', self)
        self.pushButton.clicked.connect(self.testfunc)

    def testfunc(self):
        print('hello')
        print(9 / 0)

if __name__ == '__main__':
    app = QApplication(sys.argv)
    widget = MyWidget()
    widget.show()
    app.exec_()
```

Side notes: even without `%gui qt5`, the exception is only printed if the following fix is applied: https://github.com/jupyter/qtconsole/pull/141. Also, I can't get the exception to print at all in Spyder's IPython console, after trying various settings (inline graphics vs. automatic vs. Qt5).

OS: Windows 7
Distribution: Anaconda 4.2
python: 3.5.2
ipython: 5.1.0
qtconsole: 4.2.1
pyqt: 5.6.0
jupyter: 1.0.0
spyder: 3.0.1

wmvanvliet:
IPython kernels no longer set the status='error' flag when executing a script using %run. Instead, a new message type 'error' is used for passing error messages around.

This PR adds a hacky fix to demonstrate the problem. Someone with a better grasp of the codebase than me should take a look and implement a clean solution.

fixes #156, #140

### DIFF
--- a/qtconsole/frontend_widget.py
+++ b/qtconsole/frontend_widget.py
@@ -484,8 +484,6 @@ class FrontendWidget(HistoryConsoleWidget, BaseFrontendMixin):
                 self._process_execute_ok(msg)
             elif status == 'aborted':
                 self._process_execute_abort(msg)
-            # If status == 'error', a message of type 'error' will be sent with
-            # the details
 
             self._show_interpreter_prompt_for_reply(msg)
             self.executed.emit(msg)
--- a/qtconsole/frontend_widget.py
+++ b/qtconsole/frontend_widget.py
@@ -482,7 +482,9 @@ class FrontendWidget(HistoryConsoleWidget, BaseFrontendMixin):
             status = content['status']
             if status == 'ok':
                 self._process_execute_ok(msg)
-            # Of status != 'ok', a message of type 'error' will be sent with
+            elif status == 'aborted':
+                self._process_execute_abort(msg)
+            # If status == 'error', a message of type 'error' will be sent with
             # the details
 
             self._show_interpreter_prompt_for_reply(msg)
--- a/qtconsole/frontend_widget.py
+++ b/qtconsole/frontend_widget.py
@@ -482,10 +482,8 @@ class FrontendWidget(HistoryConsoleWidget, BaseFrontendMixin):
             status = content['status']
             if status == 'ok':
                 self._process_execute_ok(msg)
-            elif status == 'error':
-                self._process_execute_error(msg)
-            elif status == 'aborted':
-                self._process_execute_abort(msg)
+            # Of status != 'ok', a message of type 'error' will be sent with
+            # the details
 
             self._show_interpreter_prompt_for_reply(msg)
             self.executed.emit(msg)
@@ -495,6 +493,11 @@ class FrontendWidget(HistoryConsoleWidget, BaseFrontendMixin):
             self._request_info['execute'].pop(msg_id)
         elif info and not self._hidden:
             raise RuntimeError("Unknown handler for %s" % info.kind)
+
+    def _handle_error(self, msg):
+        """ Handle error messages.
+        """
+        self._process_execute_error(msg)
 
     def _handle_input_request(self, msg):
         """ Handle requests for raw_input.
--- a/qtconsole/jupyter_widget.py
+++ b/qtconsole/jupyter_widget.py
@@ -293,29 +293,24 @@ class JupyterWidget(IPythonWidget):
         """Handle an execute_error message"""
         content = msg['content']
 
-        if content['ename'] == 'KeyboardInterrupt':
-            # If the user interrupted execution by pressing CTRL-C, only
-            # display a simple message.
-            self._append_plain_text("Execution aborted\n")
+        traceback = '\n'.join(content['traceback']) + '\n'
+        if False:
+            # FIXME: For now, tracebacks come as plain text, so we can't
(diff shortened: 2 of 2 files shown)


takluyver:
Thanks for working this out, I was puzzled by #140. I think that probably ipykernel should set `status='error'` in that case. Error handling is a bit weird, because we transmit the same information twice - it's something I hope to fix in a future version of the protocol.

minrk:
It's definitely a bug in the kernel if an error in a request doesn't set status=error, and that shouldn't change in the protocol. But this is the right thing to do anyway, as the error message is the right thing to use for displaying errors.

qtconsole/frontend_widget.py
```diff
@@ -482,10 +482,8 @@ def _handle_execute_reply(self, msg):
             status = content['status']
             if status == 'ok':
                 self._process_execute_ok(msg)
-            elif status == 'error':
-                self._process_execute_error(msg)
-            elif status == 'aborted':
```
takluyver: This bit for handling aborted messages probably shouldn't go missing.

wmvanvliet: aborting also triggers a 'keyboard interrupt' type error message (passed with an 'error' type message).

minrk: I think it probably should. Displaying tracebacks should only be done when `error` messages arrive.

While calling `_process_execute_aborted` makes sense to be called if it does something other than display the traceback, displaying the traceback for user code isn't part of that.

I think it's a bug that KeyboardInterrupt sets status=aborted, anyway. I'm not sure how that got in there.

minrk: https://github.com/ipython/ipykernel/issues/190 fixes the bug where KeyboardInterrupt is turned into status=aborted

tleonhardt:
This is a fix I would really like.  I am unfamiliar with how to move pull requests along for this project.  Who needs to approve the pull reqeust to merge it in?  Or is someone working on a better fix?

wmvanvliet:
a better fix requires ipython/ipykernel#190 to be merged first.

wmvanvliet:
in the meanwhile, feel free to clone this branch and use it:
`git clone https://github.com/wmvanvliet/qtconsole.git`

wmvanvliet:
Ok, I think this should do it.

In the IPython kernel, the `abort` status no longer gets set when the user presses CTRL-C (and qtconsole sends an interrupt message to the kernel). The `abort` status is now deprecated and only emitted when doing parallel computations.

When the user presses CTRL-C, this results in a plain python `KeyboardInterrupt` error message and is processed like any other error message. However, in our user interface, we want to treat this error special and don't print out a complete traceback in this case. Hence the check for `KeyboardInterrupt` is done in the UI code.

I also dropped the `ERROR: ` portion of the message, since the user requested the interrupt, so it is not an error if the execution is in fact interrupted.

qtconsole/jupyter_widget.py
```diff
@@ -292,22 +292,30 @@ def _started_channels(self):
...
-            traceback = traceback.replace(' ', '&nbsp;')
-            traceback = traceback.replace('\n', '<br/>')
-
-            ename = content['ename']
-            ename_styled = '<span class="error">%s</span>' % ename
-            traceback = traceback.replace(ename, ename_styled)
-
-            self._append_html(traceback)
+
+        if content['ename'] == 'KeyboardInterrupt':
```
takluyver: This still doesn't seem quite right to me - this is specific to a Python exception, when the Qt console can handle multiple languages. @minrk ?

minrk: No, I don't think special handling of certain Python exceptions is the right thing to do. If KeyboardInterrupt in IPython's run shouldn't display a traceback, that is probably a change that should be made in IPython, rather than the QtConsole.

wmvanvliet: I see it differently. To me, the choice whether to display a traceback or not is purely a UI decision. Hence, it should be made by the UI, not by the computing kernel.

takluyver: We want to avoid the UI making decisions based on a specific language, though, because it's supposed to work with any language. `KeyboardInterrupt` is the name of a Python exception.

wmvanvliet: I see.

takluyver:
Thanks! @ccordoba12 , want to take a

wmvanvliet:
@ccordoba12 @takluyver Gentle reminder that this PR is still waiting to be merged.

takluyver:
Thanks @wmvanvliet

Issue #122, opened 2016-04-25
QtConsole Tabs: title and motion


temok-mx:
Dear Developers,

I wonder if the QtConsole could have
-- a shortcut to move between different tabs, and
-- a command to set the title / color of the current tab.

Currently we need to use the mouse to move around 
and all tabs have titles that  makes it hard to remember 
what we are doing therein (`kernel 1`, `kernel 2`, ...).

Cheers,
Temok

wmvanvliet:
CTRL+PgUp and CTRL+PgDown move between tabs. This could be extended with a scheme like CTRL+# to jump to a specific tab. And I agree that setting the title and even color of a tab would be welcome.

tmr232:
This seems to be a duplicate of #100

wmvanvliet:
More or less, although this is about setting the title of the tabs, not the window.

tmr232:
Hmm, you're right.

PR #147, opened 2016-10-09
Added a basic hack to change the tab title via the display handlers.

### LINKED ISSUE #100: Add API to change QtConsole title
I often have multiple QtConsole windows open, and it gets very hard to tell them apart.
It would be great if there was an API to allow changing the window title.

tmr232:
You can use the following code to change a tab's title:

    def set_title(title):
        display.display({'title':title}, raw=True)

This seems like a bit of an ugly solution, abusing the display handlers, but I figured any other solution would require adding more IPC to the code so we are better off like this.

closes #100 and #122.

### DIFF
--- a/qtconsole/rich_jupyter_widget.py
+++ b/qtconsole/rich_jupyter_widget.py
@@ -164,6 +164,10 @@ class RichJupyterWidget(RichIPythonWidget):
             if 'image/svg+xml' in data:
                 svg = data['image/svg+xml']
                 self._append_svg(svg, True)
+            elif 'title' in data:
+                title = data['title']
+                tab = self.parentWidget().parentWidget()
+                tab.setTabText(tab.currentIndex(), title)
             elif 'image/png' in data:
                 # PNG data is base64 encoded as it passes over the network
                 # in a JSON structure so we decode it.


takluyver:
I think additions to the messaging protocol (which is what this is) need to be thought out more carefully. It makes sense to have something like this - we can change the title in the terminal as well, so `jupyter console` could look at the same field. I'm not sure whether it would make sense to display the 'title' somewhere in the notebook UI too.

I don't think sticking it into the `display_data` mimebundle is quite the right way to do it. A new message type would probably be the cleanest way, but we generally have quite a high bar to add message types. It could be metadata on an existing message, or a 'payload'. @minrk , thoughts?

tmr232:
I figured as much. Since this is a tiny amount of code, I thought it would not hurt to suggest it as it would not cost anyone too much time reviewing. 

For future reference - should I submit PRs to get feedback, or is there another place to bring up ideas?

takluyver:
It's fine to submit PRs to get feedback, but for things that change the messaging protocol between the kernel and the frontend, it's fairly unlikely that they'll be accepted as is. In these cases, you may want to bring it up on the [Jupyter mailing list](https://groups.google.com/forum/#!forum/jupyter) first. Or open an issue on the [jupyter_client repo](https://github.com/jupyter/jupyter_client), which is where the [message protocol docs](http://jupyter-client.readthedocs.io/en/latest/messaging.html) live.

minrk:
While I don't love payloads, this is an area where they might work. Alternately, the current trend toward custom mime-types might also allow this to be done without any extensions to the message spec.

ccordoba12:
The title of a tab is something that should be handled by the frontend, not the kernel because IMHO it doesn't have anything to do with executing code :-)

So I'd propose to close this PR and just follow #148.

PR #148, opened 2016-10-09
Added a keyboard shortcut for renaming the current tab.

### LINKED ISSUE #122: QtConsole Tabs: title and motion

tmr232:
Added the shortcut <kbd>Ctrl+R</kbd> to rename the current tab.
It also appears in the `Window` menu.

Closes #122. Relevant to #100 as well.

### DIFF
--- a/qtconsole/mainwindow.py
+++ b/qtconsole/mainwindow.py
@@ -117,7 +117,7 @@ class MainWindow(QtGui.QMainWindow):
         old_title = self.tab_widget.tabText(self.tab_widget.currentIndex())
         title, ok = QtGui.QInputDialog.getText(self,
                                                "Rename Tab",
-                                               "New Title:".format(old_title),
+                                               "New Title:",
                                                text=old_title)
         if ok:
             self.tab_widget.setTabText(self.tab_widget.currentIndex(), title)
--- a/qtconsole/mainwindow.py
+++ b/qtconsole/mainwindow.py
@@ -117,7 +117,7 @@ class MainWindow(QtGui.QMainWindow):
         old_title = self.tab_widget.tabText(self.tab_widget.currentIndex())
         title, ok = QtGui.QInputDialog.getText(self,
                                                "Rename Tab",
-                                               "New Title:",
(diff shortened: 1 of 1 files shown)


takluyver:
If we provide a way for the kernel to set the tab title as well (#147), how should they interact? Last one wins (as implemented now)? Or should the user manually setting it block changes coming from the kernel?

tmr232:
I was just about to write in #147 that on second thought, I'd be happy with this PR as a solution to #100.
Letting the kernel change the title would require a lot of UX work to figure out properly.

takluyver:
If we don't do #147 for a while, I think this is OK, but I'll give @ccordoba12 a chance to have a look at it.

qtconsole/mainwindow.py
```diff
@@ -112,6 +112,16 @@ def create_tab_with_current_kernel(self):
             name = '(%s) slave' % current_widget_name
         self.add_tab_with_frontend(widget,name=name)
 
+    def set_tab_title(self):
+        """Set the title of the current tab"""
+        old_title = self.tab_widget.tabText(self.tab_widget.currentIndex())
+        title, ok = QtGui.QInputDialog.getText(self,
+                                               "Rename Tab",
+                                               "New Title:".format(old_title),
```
wmvanvliet: we don't need the `.format(old_title)` here

wmvanvliet: New Title -> New title

tmr232: Fixed.

wmvanvliet: I still think the label should read "New title:" and not "New Title", there is no reason to use title casing here.

ccordoba12:
+1 from me :-)

PR #152, opened 2016-10-13
Create new tab with existing kernel

### LINKED ISSUE #101: Create new tab with specified existing kernel
I want to be able to create a new tab and connect it to an existing kernel.
Searching for a solution, the only thing I found is [this](http://stackoverflow.com/questions/10496562/is-it-possible-to-connect-to-several-backends-with-one-ipython-qtconsole-fronten) StackOverflow thread.

tmr232:
Fixes #101 

The code does what is should, but I have 2 issues that need to be addressed:
1. Choosing the kernel. Currently you just enter the kernel name (from the `--existing` option). We might want something more user friendly,
2. Naming the new tab - should it be automatic ("external 0", "external 1" and so on) or user defined?

### DIFF
--- a/qtconsole/mainwindow.py
+++ b/qtconsole/mainwindow.py
@@ -6,10 +6,12 @@ common actions.
 
 # Copyright (c) Jupyter Development Team.
 # Distributed under the terms of the Modified BSD License.
-
+import os
 import sys
 import webbrowser
 from threading import Thread
+
+from jupyter_core.paths import jupyter_runtime_dir
 
 from qtconsole.qt import QtGui,QtCore
 from qtconsole.usage import gui_reference
@@ -31,6 +33,7 @@ class MainWindow(QtGui.QMainWindow):
     def __init__(self, app,
                     confirm_exit=True,
                     new_frontend_factory=None, slave_frontend_factory=None,
+                    connection_frontend_factory=None,
                 ):
         """ Create a tabbed MainWindow for managing FrontendWidgets
         
@@ -54,6 +57,7 @@ class MainWindow(QtGui.QMainWindow):
         self.confirm_exit = confirm_exit
         self.new_frontend_factory = new_frontend_factory
         self.slave_frontend_factory = slave_frontend_factory
+        self.connection_frontend_factory = connection_frontend_factory
 
         self.tab_widget = QtGui.QTabWidget(self)
         self.tab_widget.setDocumentMode(True)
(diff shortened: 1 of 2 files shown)


qtconsole/mainwindow.py
```diff
@@ -109,6 +113,17 @@ def set_window_title(self):
         if ok:
             self.setWindowTitle(title)
 
+    def create_tab_with_existing_kernel(self):
+        name, ok = QtGui.QInputDialog.getText(self,
+                                               "Connect to Existing Kernel",
+                                               "Kernel name:")
+        if not ok:
+            return
+        connection_file = os.path.join(jupyter_runtime_dir(), name)
```
ccordoba12: There's a function called `find_connection_file` from `jupyter_client` (I think) that you should use here

qtconsole/mainwindow.py
```diff
@@ -109,6 +113,17 @@ def set_window_title(self):
...
+        widget = self.connection_frontend_factory(connection_file)
+        name = 'bla'
```
ccordoba12: ??

tmr232: As I said in the pull request mesaage - I consider tab naming an open issue. So I figured I should get some input before writing throw-away code.

ccordoba12:
About your points:

1. You could offer a QfileDialog to let users select a kernel.json file from the OS.
2. I think naming should be automatic.

qtconsole/mainwindow.py
tmr232: In the new version it is not used as I use the file dialog to find the connection files. The `jupyter_runtime_dir()` function is now needed to find the actual directory.

tmr232:
@ccordoba12 anything else needed before this can be merged?

qtconsole/qtconsoleapp.py
```diff
@@ -206,6 +206,31 @@ def new_frontend_master(self):
...
         return widget
 
+    def new_frontend_connection(self, connection_file):
+        """Create and return a new frontend attached to an existing kernel.
+
+                Parameters
+                ----------
+                connection_file : str
+                    The connection_file path this frontend is to connect to
+                """
```
ccordoba12: Please align this docstring correctly

ccordoba12:
I left a minor comment, but otherwise this looks to me.

ccordoba12:
LGTM, thanks!

Issue #157, opened 2016-11-04
display jpg image doesn't work


mickeydonkey:
```
from IPython.core.display import *
Image(data='https://upload.wikimedia.org/wikipedia/commons/thumb/2/29/Pomegranate03_edit.jpg/320px-Pomegranate03_edit.jpg', embed=True,format='jpeg')

Out[16]: <IPython.core.display.Image object>
```
qtconsole=4.2.1
python=3.5
verified on both windows and Ubuntu 16.04

@ivanov further tested in python2, seem it is relevant to python3 only.

mickeydonkey:
Since QtConsole already support matplotlib with SVG as backend, I believe It's a good idea to support SVG display.

takluyver:
#158 fixes it for me

mickeydonkey:
@takluyver Very fast fix, Thank you! what's your opinion about SVG display?

mickeydonkey:
`Image('/path/to/pattern.svg')` gives an error `Cannot embed the 'svg' image format` 
I've managed to display SVG in qtconsole, but It seems not the first-class citizen like png and jpeg.
```
svg = open('/home/luot/Pictures/pattern.svg', 'r').read()
display_svg(svg, raw=True)
```

mickeydonkey:
seems https://github.com/ipython/ipython/issues/10043 is the right place for SVG display support.

dashesy:
I have this problem in `Jupyter QtConsole 4.2.1` and was wondering if it is fixed in QtConsole 4.3

wmvanvliet:
it should be. #158 was merged before the 4.3 release.

PR #158, opened 2016-11-04
Fix JPEG support check in Python 3

takluyver:
Closes gh-157

Is this the last bug that has managed to survive since the Python 3 migration? ;-)


ivanov:
nice work! 🍰

PR #164, opened 2016-11-18
Stop special sequences from overriding custom keyboard shortcuts

wmvanvliet:
On OSX, Ctrl+K is mapped to 'cut', Ctrl+Y is mapped to 'paste', etc. These interfere with the readline-like binding we'd like to support on this platform (e.g. Ctrl+K should delete everything until the end of line). This PR shuffles the priority of the keyboard shortcuts, so that the readline-like bindings get priority.


ccordoba12:
LGTM, thanks.

PR #173, opened 2016-12-09
Make browse history to work for lines longer than the console's width

mariacamilaremolinagutierrez:
Fixes https://github.com/spyder-ide/spyder/issues/3215

### DIFF
--- a/qtconsole/console_widget.py
+++ b/qtconsole/console_widget.py
@@ -1591,17 +1591,6 @@ class ConsoleWidget(MetaQObjectHasTraits('NewBase', (LoggingConfigurable, superQ
             cursor = self._control.textCursor()
             return cursor.columnNumber() - len(prompt)
 
-    def _get_input_buffer_cursor_position_in_block(self):
-        """ Returns the position of the cursor in the block. Returns -1 if 
-            there is no prompt.
-        """
-        prompt = self._get_input_buffer_cursor_prompt()
-        if prompt is None:
-            return -1
-        else:
-            cursor = self._control.textCursor()
-            return cursor.positionInBlock()
-
     def _get_prompt_position(self):
         """ Returns the position of the prompt.
         """
--- a/qtconsole/console_widget.py
+++ b/qtconsole/console_widget.py
@@ -1591,6 +1591,26 @@ class ConsoleWidget(MetaQObjectHasTraits('NewBase', (LoggingConfigurable, superQ
             cursor = self._control.textCursor()
             return cursor.columnNumber() - len(prompt)
 
+    def _get_input_buffer_cursor_position_in_block(self):
+        """ Returns the position of the cursor in the block. Returns -1 if 
+            there is no prompt.
+        """
+        prompt = self._get_input_buffer_cursor_prompt()
+        if prompt is None:
+            return -1
+        else:
+            cursor = self._control.textCursor()
+            return cursor.positionInBlock()
+
+    def _get_prompt_position(self):
+        """ Returns the position of the prompt.
+        """
+        prompt = self._get_input_buffer_cursor_prompt()
+        if prompt is None:
+            return -1
+        else:
+            return len(prompt)
+
     def _get_input_buffer_cursor_line(self):
         """ Returns the text of the line of the input buffer that contains the
             cursor, or None if there is no such line.
--- a/qtconsole/history_console_widget.py
+++ b/qtconsole/history_console_widget.py
@@ -76,7 +76,7 @@ class HistoryConsoleWidget(ConsoleWidget):
                 return False
 
             # Set a search prefix based on the cursor position.
-            pos = self._get_input_buffer_cursor_position_in_block()
+            pos = self._get_input_buffer_cursor_pos()
             input_buffer = self.input_buffer
             # use the *shortest* of the cursor column and the history prefix
             # to determine if the prefix has changed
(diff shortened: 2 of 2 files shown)


qtconsole/history_console_widget.py
```diff
@@ -76,31 +76,36 @@ def _up_pressed(self, shift_modifier):
...
             # and the cursor is at the end of the first line
-            
+
             # check if we are at the end of the first line
             c = self._get_cursor()
             current_pos = c.position()
-            c.movePosition(QtGui.QTextCursor.EndOfLine)
-            at_eol = (c.position() == current_pos)
-            
+            if self.console_width > pos:
```
minrk: `self.console_width` isn't a trustworthy metric. This is an input value for setting the initial width of the window, not the current window size.

qtconsole/history_console_widget.py
```diff
@@ -76,31 +76,36 @@ def _up_pressed(self, shift_modifier):
...
+                c.movePosition(QtGui.QTextCursor.EndOfBlock)
+                at_eol = (c.position() == current_pos)
+            else:
+                c.movePosition(QtGui.QTextCursor.EndOfBlock)
+                col = c.columnNumber()
+                at_eol = (c.position() == current_pos + col)
```
minrk: I don't think this block is quite correct. When I am at the end of a wrapped line, I get that eol should be `c.position() == current_pos`, not `current_pos + col`. I think you can do without the `else:` block here, and just make the change to use `EndOfBlock` instead of `EndOfLine`.

qtconsole/history_console_widget.py
```diff
@@ -76,31 +76,36 @@ def _up_pressed(self, shift_modifier):
...
-            
+
             # check if we are at the end of the first line
             c = self._get_cursor()
             current_pos = c.position()
-            c.movePosition(QtGui.QTextCursor.EndOfLine)
-            at_eol = (c.position() == current_pos)
-            
+            if self.console_width > pos:
+                c.movePosition(QtGui.QTextCursor.EndOfBlock)
```
minrk: Since you've changed this to `EndOfBlock`, I think you need to change the `cursor.movePosition(QtGui.QTextCursor.EndOfLine)` below to `EndOfBlock` as well.

mariacamilaremolinagutierrez:
@minrk thank you very much for your revision, I completely agree with you. Last night I was getting some weird errors and that was why I made so many tricks, but it is working now because of a dumb bug in my previous implementation. I think this new commit goes according to your comments :)

wmvanvliet:
For me on OSX, I can reproduce the history problem mentioned in spyder-ide/spyder#3215, but this PR in it's current form breaks it even further. Now whenever there is a line that spans multiple lines, the up and down keys stop working.

mariacamilaremolinagutierrez:
@wmvanvliet You're right. And this bugs me a lot because I thought it first the way it is now and it didn't work but then it did (I'm sure it was my mistake, but it still inda weird). I tested it with my initial commit (https://github.com/mariacamilaremolinagutierrez/qtconsole/blob/9d74bca1bb10f28b9f16b2c88ece340a31c1f4f8/qtconsole/history_console_widget.py) and it is working. I know this still has the problem that ```self.console_width``` isn't a trustworthy metric, but that can be fixed asking the current width not the initial one. 

Does this commit work for you? (you only need to use the file I'm linking)

Also thanks for reviewing this :)

wmvanvliet:
if I have done my git commands correctly... no, also the first commit does not work for me on OSX.

qtconsole/console_widget.py
```diff
@@ -1591,6 +1591,17 @@ def _get_input_buffer_cursor_column(self):
...
+        """ Returns the position of the cursor in the block. Returns -1 if 
+            there is no prompt.
+        """
+        prompt = self._get_input_buffer_cursor_prompt()
+        if prompt is None:
+            return -1
+        else:
+            cursor = self._control.textCursor()
+            return cursor.positionInBlock()
+
```
wmvanvliet: You don't need this function.

qtconsole/history_console_widget.py
```diff
@@ -76,31 +76,31 @@ def _up_pressed(self, shift_modifier):
...
             # to the line up to the cursor is if we are already
             # in a simple scroll (no prefix),
             # and the cursor is at the end of the first line
-            
+
             # check if we are at the end of the first line
             c = self._get_cursor()
             current_pos = c.position()
-            c.movePosition(QtGui.QTextCursor.EndOfLine)
+            c.movePosition(QtGui.QTextCursor.EndOfBlock)
```
wmvanvliet: also change line 116/117 to: `cursor.movePosition(QtGui.QTextCuresor.EndOfBlock)`

qtconsole/history_console_widget.py
```diff
@@ -76,31 +76,31 @@ def _up_pressed(self, shift_modifier):
                 return False
 
             # Set a search prefix based on the cursor position.
-            col = self._get_input_buffer_cursor_column()
+            pos = self._get_input_buffer_cursor_position_in_block()
```
wmvanvliet: use: `pos = self._get_input_buffer_cursor_pos()`

mariacamilaremolinagutierrez:
@wmvanvliet thank you very much for your help in this, it is working for me as well 😄 I just pushed the changes.

ccordoba12:
Thanks @minrk and @wmvanvliet for your reviews, and of course @mariacamilaremolinagutierrez for your work :-)

minrk:
Awesome, thanks everyone!

Issue #168, opened 2016-12-01
Unable to open qtconsole - RuntimeError


chrisjcameron:
In Python 3.5 I am having an issue launching qtconsole. imp.py expects the `mod.__path__` to be of type list but `__import__` is packaging the `__path__` inside a `_NamespacePath` with a `._path` attribute. I suspect a lot of the issue is using a depreciated module (`imp.py`). 

I was able to workaround in `qt_loaders.py` by modifying `has_binding(api)` to detect namespace objects and send the expected input to `find_module`:
Line 126:

```
        path = mod.__path__
        if hasattr(mod.__path__, '_path' ):
            path = mod.__path__._path
        imp.find_module('QtCore', path )
        imp.find_module('QtGui', path)
        imp.find_module('QtSvg', path)
        if api == QT_API_PYQT5:
            # QT5 requires QtWidgets too
            imp.find_module('QtWidgets', path)

```

The error: 

```
bureaucratic-node-160$ jupyter qtconsole
_NamespacePath(['/Users/cjc73/anaconda/lib/python3.5/site-packages/PyQt5'])
Traceback (most recent call last):
  File "/Users/cjc73/anaconda/bin/jupyter-qtconsole", line 2, in <module>
    from qtconsole.qtconsoleapp import main
  File "/Users/cjc73/anaconda/lib/python3.5/site-packages/qtconsole/qtconsoleapp.py", line 60, in <module>
    from qtconsole.qt import QtCore, QtGui
  File "/Users/cjc73/anaconda/lib/python3.5/site-packages/qtconsole/qt.py", line 23, in <module>
    QtCore, QtGui, QtSvg, QT_API = load_qt(api_opts)
  File "/Users/cjc73/anaconda/lib/python3.5/site-packages/qtconsole/qt_loaders.py", line 286, in load_qt
    if not can_import(api):
  File "/Users/cjc73/anaconda/lib/python3.5/site-packages/qtconsole/qt_loaders.py", line 161, in can_import
    if not has_binding(api):
  File "/Users/cjc73/anaconda/lib/python3.5/site-packages/qtconsole/qt_loaders.py", line 126, in has_binding
    imp.find_module('QtCore', mod.__path__)
  File "/Users/cjc73/anaconda/lib/python3.5/imp.py", line 271, in find_module
    "not {}".format(type(name)))
RuntimeError: 'list' must be None or a list, not <class 'str'>
```

takluyver:
[...]

PR #183, opened 2017-01-12
Fix PyQt import checks on Python 3.5

takluyver:
This provides a new implementation for Python >= 3.4, leaving the existing one for Python <= 3.3.

Closes gh-168 (hopefully; I can't reproduce it to test)


takluyver:
Now that I've done this, I see that #139 had already fixed the error. I think this is still worth adding, as it uses `importlib` where possible in place of the deprecated `imp` module. But it's lower priority.

PR #184, opened 2017-01-13
[WIP] Switching syntax styles option

mariacamilaremolinagutierrez:
Now there is an option to change to a different syntax style that opens a new tab in the console with the selected color scheme:

image1

### DIFF
--- a/qtconsole/mainwindow.py
+++ b/qtconsole/mainwindow.py
@@ -740,14 +740,9 @@ class MainWindow(QtGui.QMainWindow):
         self.active_frontend._set_paging(paging)
 
     def get_available_syntax_styles(self):
-        import pygments, os
-        syntax_path = pygments.__file__.rpartition('/')[0] + '/styles/'
-        directory_elements = os.listdir(syntax_path)
-        available_syntax_styles = []
-        for de in directory_elements:
-            if os.path.isfile(syntax_path+de) and de != '__init__.py':
-                available_syntax_styles.append(de.split('.py')[0])
-        return sorted(available_syntax_styles)
+        from pygments.styles import get_all_styles
+        styles = list(get_all_styles())
+        return sorted(styles)
 
     def set_syntax_style(self, syntax_style):
         widget = self.new_frontend_factory(syntax_style=syntax_style)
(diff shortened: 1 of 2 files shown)


mariacamilaremolinagutierrez:
@ccordoba12 @minrk I was doing this PR in order to find out if there was a way of changing the syntax_style without creating a new tab but that instead it got updated automatically in the same shell I've been working on.

The initial intention was inspired by spyder's https://github.com/spyder-ide/spyder/pull/3908. However I was unable to do so. Do any of you know if this is possible? Or do I have to restart it and initialize it with the new config (as in this PR)?

takluyver:
`JupyterWidget` has a `_syntax_style_changed()` method, which should be called automatically when `syntax_style` is set. So if you can get the active JupyterWidget instance, you should be able to do:

```python
jupyter_widget.syntax_style = new_style
```

It might be that that doesn't work, though, if no-one's tested it before.

qtconsole/mainwindow.py
```diff
@@ -730,6 +739,20 @@ def toggleFullScreen(self):
     def set_paging_active_frontend(self, paging):
         self.active_frontend._set_paging(paging)
 
+    def get_available_syntax_styles(self):
```
takluyver: We can use a pygments function to find the available styles:
http://pygments.org/docs/styles/#getting-a-list-of-available-styles

wmvanvliet:
awesome!

qtconsole/mainwindow.py
mariacamilaremolinagutierrez: Oh this is way better! Thank you!

mariacamilaremolinagutierrez:
> JupyterWidget has a ```_syntax_style_changed()``` method, which should be called automatically when ```syntax_style``` is set. So if you can get the active JupyterWidget instance, you should be able to do: ```jupyter_widget.syntax_style = new_style```

@takluyver I tried to do so before in another project (Spyder IDE) where I have the JupyterWidget instance: 

image

The interesting thing was that the style_sheet was updated automatically (without restarting it or making a new tab). But the syntax style did not, it only got updated after I restarted Spyder. 

This makes me think about your last line and that it maybe doesn't work. I tried to check both functions a bit but I couldn't figure out fast what could be wrong with the ```_syntax_style_changed()``` one.

wmvanvliet:
Since this creates a new kernel, shouldn't this option do in the "Kernel" menu? Is there no way to change the pygments coloring of the currently running kernel? That would be super.

ccordoba12:
@wmvanvliet, we'll fix that because we need it for Spyder.

ccordoba12:
Closing in favor of #217.

PR #189, opened 2017-02-06
Fix bug where pasting does not work properly

wmvanvliet:
#164 broke multiline pasting on Linux (and probably Windows). This PR provides a fix.


ccordoba12:
Could you add a test for this?

wmvanvliet:
at some point, there should be a comprehensive test for all the keybindings. For now, I've added just the scenario that we were trying to fix.

ccordoba12:
Great!! Thanks a lot for taking the extra effort!

PR #196, opened 2017-03-20
Enhance handling of input from other clients

wmvanvliet:
The `c.ConsoleWidget.include_other_output = True` switch will make QtConsole display input given to the kernel that was initiated by other clients. For example, a text editor can connect to the kernel and allow users to select pieces of code and execute them (like the F9 button in Spyder and MATLAB).

This PR fixes some bugs and adds some enhancements when displaying "other_input" and "other_output".

 1. Adds an `[remote] In [n]: ` prompt for other_input and an `[remote] Out[n]: ` prompt for other_output (like jupyter-console does)
 2. Adds a configuration option `other_output_prefix` to change the `[remote] ` prefix above to something else (like jupyter-console has)
 3. Fixes the syntax highlighting of other_input. The decision on whether to highlight a block is made based on whether the block starts with a prompt. The prompt detection code has been updated to properly detect the prompt generated by other_input.
 4. After other_input is received, it includes an updated prompt number. The number of the current prompt is updated in-place to reflect this.

### DIFF
--- a/qtconsole/console_widget.py
+++ b/qtconsole/console_widget.py
@@ -495,6 +495,13 @@ class ConsoleWidget(MetaQObjectHasTraits('NewBase', (LoggingConfigurable, superQ
         Outputs are not displayed until enter is pressed.
         """
     )
+
+    other_output_prefix = Unicode('[remote] ', config=True,
+        help="""Prefix to add to outputs coming from clients other than this one.
+
+        Only relevant if include_other_output is True.
+        """
+    )
     
     def can_copy(self):
         """ Returns whether text can be copied to the clipboard.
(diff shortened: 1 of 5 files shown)


ccordoba12:
> This PR gives displaying whatever text I send to the kernel, some much needed love.

What were the previous failures that this PR fixes?

qtconsole/console_widget.py
```diff
@@ -494,6 +494,13 @@ def sizeHint(self):
         Outputs are not displayed until enter is pressed.
         """
     )
+
+    other_output_prefix = Unicode('[remote] ', config=True,
```
ccordoba12: Why is this value set to `[remote]`?

wmvanvliet: That's the default used in the jupyter console client. I copied the feature from there.

wmvanvliet:
I've updated the description of this PR to be more informative.

qtconsole/console_widget.py
ccordoba12: Ok, didn't know that.

PR #198, opened 2017-03-22
Fix saving HTML with inline PNG for python3

### LINKED ISSUE #51: HTML with inline png images does not appear to be generated correctly
To reproduce:
1.  From within Spyder, Tools->Preferences->IPython console->Graphics, set the Graphics backend to “Inline”.
2.  Close Spyder and restart.
3.  Run this to display an inline plot (for example):
   
   ```
                  import matplotlib.pyplot as plt
                  plt.plot([1, 3, 1, 6, 10, 4])
   ```
4.  Hit control-s, choose a file name and save as “HTML with PNG figures”, click “save”.
5.  Select “Inline” when it asks “Would you like inline PNGs …” 
6.  Neither FireFox nor IE will display the figure.

The HTML ends up with code like this:

```
     <img src="data:image/png;base64,b'iVBORw0KGgoAAAAN...'
```

To get the figure to display, I manually deleted the leading b' and the trailing '.

I finally tracked the source of this down to the file ‘rich_jupyter_widget.py’, in the routine ‘RichJupyterWidget._get_image_tag()’.  The offending code is:

```
    re.sub(r'(.{60})',r'\1\n',str(ba.toBase64()))
```

Modifying this to:

```
    re.sub(r'(.{60})',r'\1\n',str(ba.toBase64())[2:-1])
```

works, but I don’t know the what the best fix is or if this causes other issues somewhere else.

wmvanvliet:
When using python 3, saving to HTML with inline PNG images is reported broken (#51). This fixes it.

fixes #51

### DIFF
--- a/qtconsole/rich_jupyter_widget.py
+++ b/qtconsole/rich_jupyter_widget.py
@@ -304,7 +304,7 @@ class RichJupyterWidget(RichIPythonWidget):
                 image.save(buffer_, format.upper())
                 buffer_.close()
                 return '<img src="data:image/%s;base64,\n%s\n" />' % (
-                    format,re.sub(r'(.{60})',r'\1\n',str(ba.toBase64())))
+                    format,re.sub(r'(.{60})',r'\1\n', str(ba.toBase64().data().decode())))
 
         elif format == "svg":
             try:


ccordoba12:
Does this work for Python 2 as well?

ccordoba12:
Sorry, I just saw that the OP of #51 reported that it works for both.

Merging then.

PR #200, opened 2017-03-23
Change warning to a note in changelog

willingc:
@ccordoba12 Here's an attempt to make the warning less scary for those considering upgrade.


ccordoba12:
Looks good to me. Thanks @willingc!

PR #242, opened 2017-10-25
Make behavior of Ctrl+Backspace and Ctrl+Delete more intelligent

### LINKED ISSUE #241: Ctrl-Backspace behaves inconsistently in iPython qtconsole (within Spyder)
I first submitted this over on the Spyder repo, but was helpfully informed the issue lies on this end. Thus:

## Description

Often, when sending or copying code to the console, I might want to delete a blank line or clean it up to all be on one line. However, unlike in the Spyder editor and most all other editors/consoles, Ctrl-Backspace doesn't just delete extra spaces and such from the current cursor position to the beginning of the line, but also the line break and the last word on the previous line. 

In the (perhaps uncommon) case one might actually want to delete either of these as well, another one or two taps on backspace key will do the trick, but conversely there is no easy way to replicate the other behavior, and given the inconsistency with the editor and most other applications it often leads to mistaken deletions. 

Therefore, the newline character and the blank spaces in a line should each be treated as separate words for the purposes of Ctrl-Backspace, as they are elsewhere.

**What steps will reproduce the problem?**

1. Copy or type a multiline statement into the console, e.g.
```
foo = ["foo", "foo", "foo",
          "foo", foo", "bar"]
```
2. Place cursor at beginning of second line text (e.g. with Home)
3. Hit Ctrl-Backspace

**What is the expected output? What do you see instead?**

Expected: Blank spaces are deleted, up to the beginning of the line, as in the Spyder editor and most other decent consoles
e.g.
```
foo = ["foo", "foo", "foo",
"foo", foo", "bar"]
```
and then an additional tap of Ctrl-Backspace gives
```
foo = ["foo", "foo", "foo","foo", foo", "bar"]
```
with further such keystrokes deleting successive "word" elements as normal.

Instead, with only one press of Ctrl-Backspace, I get
```
foo = ["foo", "foo", ""foo", foo", "bar"]
```
i.e. the blank spaces, the line break, and the last "word" element all deleted.

## Version and main components

* qtconsole 4.3.1
* IPython 6.1.0
* Spyder Version: 3.2.4 
* Python Version: 3.6.3
* Qt Versions: 5.6.2, PyQt5 5.6 on Windows 8.1 x64

sys_info():
```
{'commit_hash': 'd86648c5d',
 'commit_source': 'installation',
 'default_encoding': 'utf-8',
 'ipython_path': 'C:\\Anaconda3\\lib\\site-packages\\IPython',
 'ipython_version': '6.1.0',
 'os_name': 'nt',
 'platform': 'Windows-8.1-6.3.9600-SP0',
 'sys_executable': 'C:\\Anaconda3\\python.exe',
 'sys_platform': 'win32',
 'sys_version': '3.6.3 |Anaconda, Inc.| (default, Oct 15 2017, 03:27:45) [MSCv.1900 64 bit (AMD64)]'}
```

## Spyder/Jupyter Dependencies
```
pyflakes >=0.6.0 :  1.6.0 (OK)
pycodestyle >=2.3:  2.3.1 (OK)
pygments >=2.0   :  2.2.0 (OK)
pandas >=0.13.1  :  0.20.3 (OK)
numpy >=1.7      :  1.13.3 (OK)
sphinx >=0.6.6   :  1.6.3 (OK)
rope >=0.9.4     :  0.10.5 (OK)
jedi >=0.9.0     :  0.10.2 (OK)
nbconvert >=4.0  :  5.3.1 (OK)
sympy >=0.7.3    :  1.1.1 (OK)
cython >=0.21    :  0.26.1 (OK)
qtconsole >=4.2.0:  4.3.1 (OK)
IPython >=4.0    :  6.1.0 (OK)
pylint >=0.25    :  1.7.4 (OK)

```

wmvanvliet:
This stops them from deleting too greedily across line boundaries.

    > foo = ['foo', 'foo', 'foo',
    >       'bar']

One Ctrl+Backspace will do:

    > foo = ['foo', 'foo', 'foo',
    >       '

Another will do:

    > foo = ['foo', 'foo', 'foo',
    >

And another will do:

    > foo = ['foo', 'foo', 'foo',

Also PEP8-i-fied some related code.

Fixes #241

### DIFF
--- a/qtconsole/tests/test_console_widget.py
+++ b/qtconsole/tests/test_console_widget.py
@@ -151,7 +151,7 @@ class TestConsoleWidget(unittest.TestCase):
         c = control.textCursor()
         c.setPosition(c.position() - 3)
         control.setTextCursor(c)
-        QTest.keyClick(control, QtCore.Qt.Key_K, QtCore.Qt.MetaModifier)
+        QTest.keyClick(control, QtCore.Qt.Key_K, QtCore.Qt.ControlModifier)
         self.assertEqual(w._get_input_buffer(), 'test in')
 
         # Ctrl+V pastes
--- a/qtconsole/tests/test_console_widget.py
+++ b/qtconsole/tests/test_console_widget.py
@@ -142,9 +142,6 @@ class TestConsoleWidget(unittest.TestCase):
         w._show_prompt()
         control = w._control
 
-        print(w._continuation_prompt +'|')
-        print(w._continuation_prompt_html)
-
         # Test setting the input buffer
         w._set_input_buffer('test input')
         self.assertEqual(w._get_input_buffer(), 'test input')
@@ -154,7 +151,7 @@ class TestConsoleWidget(unittest.TestCase):
         c = control.textCursor()
         c.setPosition(c.position() - 3)
         control.setTextCursor(c)
-        QTest.keyClick(control, QtCore.Qt.Key_K, QtCore.Qt.ControlModifier)
+        QTest.keyClick(control, QtCore.Qt.Key_K, QtCore.Qt.MetaModifier)
         self.assertEqual(w._get_input_buffer(), 'test in')
 
         # Ctrl+V pastes
@@ -189,24 +186,24 @@ class TestConsoleWidget(unittest.TestCase):
         w._set_input_buffer("foo = ['foo', 'foo', 'foo',    \n"
                             "       'bar', 'bar', 'bar']")
         QTest.keyClick(control, QtCore.Qt.Key_Backspace,
-                       QtCore.Qt.ControlModifier)
+                       QtCore.Qt.MetaModifier)
         self.assertEqual(w._get_input_buffer(),
                          ("foo = ['foo', 'foo', 'foo',    \n"
                             "       'bar', 'bar', '"))
         QTest.keyClick(control, QtCore.Qt.Key_Backspace,
-                       QtCore.Qt.ControlModifier)
+                       QtCore.Qt.MetaModifier)
         QTest.keyClick(control, QtCore.Qt.Key_Backspace,
-                       QtCore.Qt.ControlModifier)
+                       QtCore.Qt.MetaModifier)
         self.assertEqual(w._get_input_buffer(),
                          ("foo = ['foo', 'foo', 'foo',    \n"
                           "       '"))
         QTest.keyClick(control, QtCore.Qt.Key_Backspace,
-                       QtCore.Qt.ControlModifier)
+                       QtCore.Qt.MetaModifier)
         self.assertEqual(w._get_input_buffer(),
                          ("foo = ['foo', 'foo', 'foo',    \n"
                           ""))
         QTest.keyClick(control, QtCore.Qt.Key_Backspace,
-                       QtCore.Qt.ControlModifier)
+                       QtCore.Qt.MetaModifier)
         self.assertEqual(w._get_input_buffer(),
                          "foo = ['foo', 'foo', 'foo',")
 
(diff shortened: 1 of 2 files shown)


qtconsole/tests/test_console_widget.py
```diff
@@ -138,6 +142,9 @@ def test_keypresses(self):
         w._show_prompt()
         control = w._control
 
+        print(w._continuation_prompt +'|')
+        print(w._continuation_prompt_html)
```
ccordoba12: Are these print's necessary?

CAMGerlach-Baron:
@ccordoba12 @wmvanvliet My apologies for the delay; took a while to get a proper dev environment setup for this since I'm new to Python and am an atmospheric science grad student without a formal programming background. 

I can confirm it indeed solves the immediate issue as directly stated in #241 ; going the extra mile to standardize `Ctrl-Del` as well from the end of a line. (qt)Awesome! As a sidenote (I can file a new issue over there) Spyder actually has slightly nonstandard handling of the latter—the common, and probably most sensible convention is for the first `Ctrl-Del` to remove any whitespace at the end of the line, and the second (or the first, if no whitespace existed) to both delete the linebreak and any now-redundant indents—this is what your PR branch now properly does, while Spyder only deletes the linebreak and requires another tap to nuke the spaces. A very minor issue for sure, but consistency is always nice.

As a corollary, a directly related inconsistency still exists in this build—in Spyder and most other editors, `Ctrl-Del` before some whitespace and then a word (beginning of the line or not) deletes only the whitespace with the first tap, rather than the whitespace + word as `qtconsole` does. The former is consistent with the standard handling of `Ctrl-rightarrow`, including in `qtconsole`, and allows for a quick `Ctrl-A`, `Ctrl-Del` in `qtconsole` to delete all whitespace/indents at the beginning of a line. So, if the cursor ("`I`") is e.g. at line 2, col 3:
```
do_something("foo", "bar",
  I                     "whatsit")
```
Expected:
```
do_something("foo", "bar",
  I"whatsit")
```
Actual:
```
do_something("foo", "bar",
  I")
```

Also, while testing noticed seemingly odd behavior (in this build and master) where hitting `rightarrow` at the end of a line of a multiline statement would move the cursor all the way to the end of the last line of the statement, rather than the beginning of the next line; to get to the latter one must do `downarrow`, `home`. Feature/consistent with some other convention? Or should I file an issue? Thanks!

wmvanvliet:
The behaviour of the right arrow is weird indeed and is probably a bug.

CAMGerlach-Baron:
Okay, thanks @wmvanvliet . Should I file a separate issue then?

wmvanvliet:
Yes. Please do. And can you check the latest commit and see if that fixes the behaviour of CTRL+DEL for you?

CAMGerlach-Baron:
Bad news—Ctrl-Backspace still works, as does regular delete, but on the same Windows 10 Ent x64 machine and conda env I tested it previously, one tap of Ctrl-Delete causes qtconsoleapp and the python session that spawned it to hang, which must then be interrupted with Ctrl-C which kills the python interpreter, or python killed from the OS which obviously does the same.

wmvanvliet:
Try it now :)

CAMGerlach-Baron:
Works perfectly; thanks so much! 

(Also, filed that other bug, per your suggestion)

wmvanvliet:
Allright. Then I guess this PR is ready for review.

ccordoba12:
@wmvanvliet, I don't have time to review this one. If you feel confident about the code, please merge.

I don't know if @takluyver would like to chime in, though.

takluyver:
I've had a brief scan through the code, but I don't think about this code often enough to anticipate potential problems. It looks clear, and you've added some tests for the behaviour of Ctrl-Backspace and Ctrl-Delete, so I'd say go ahead and merge it. Thanks!

Issue #243, opened 2017-10-31
Is it possible to connect to other kernels ?


daegontaven:
Hi, I'm building an application that embeds an in-process qtconsole. I was wondering If i can connect QtConsole to IRuby or any kernels of my choosing. Could you provide any documentation for it ?

daegontaven:
@minrk Could you please look into this ? According to the example provided:
https://github.com/jupyter/qtconsole/blob/6b099f7f128766ec905d5a8dd18dba763d4ac54a/examples/inprocess_qtconsole.py#L20-L23

`QtInProcessKernelManager()` seems to be a wrapper around `ipykernel` of IPython. It's already possible to connect to other kernels(eg for IRPython `jupyter qtconsole --kernel=ir`) from jupyter. My question is whether it can be done using in-process. Perhaps I could launch a subprocess of IRPython's kernel from my PyQt application and connect to it using methods provided by `RichJupyterWidget()`. However, I can't seem to find any public methods in that class. Also more worrying is the fact that it is inheriting from an `IpythonWidget` dummy class.

Does this imply there is no current in-process support for qtconsole ? If not, how can I start contributing to make it more general purpose ?

Thank you in advance ✌️

takluyver:
You cannot run another kernel in-process (to be honest it doesn't work well even for Python). The supported way to use kernels is for them to be in a separate process.

daegontaven:
@takluyver are there example on how to do it as a separate process(I think i meant separate process instead of in-process)? I see no examples in the documentation given [here](https://qtconsole.readthedocs.io/en/stable/#embedding-the-qtconsole-in-a-qt-application). How would I connect to a separate kernel ? How would I start the kernel(subprocess using jupyter command ?) ? Can you point to me to any information ?

takluyver:
There should be examples, but there aren't, unfortunately. And I don't actually know how to do it. @ccordoba12 uses the Qt console embedded in Spyder, he might have a better idea what's needed.

ccordoba12:
We have our own implementation of `qtconsole` and it's a complicated setup, but here it goes:

1. We have our own kernel spec to start our kernels:

    https://github.com/spyder-ide/spyder/blob/3.x/spyder/utils/ipython/kernelspec.py

2. We use it to create our kernel managers and clients, in the `create_kernel_manager_and_kernel_client` method of

    https://github.com/spyder-ide/spyder/blob/3.x/spyder/plugins/ipythonconsole.py

[...]

PR #244, opened 2017-11-01
Create a simple example of embedding, and clarify docs.

takluyver:
Starting the kernel in a separate process should be the default option for most users.

Closes gh-243

### DIFF
--- a/docs/source/index.rst
+++ b/docs/source/index.rst
@@ -445,31 +445,28 @@ garbage collected until the application itself is destroyed.
 Embedding the QtConsole in a Qt application
 -------------------------------------------
 
-In order to make the QtConsole available to an external Qt GUI application (just as
-:func:`IPython.embed` enables one to embed a terminal session of IPython in a
-command-line application), there are a few options:
-
-* First start IPython, and then start the external Qt application from IPython,
-  as described above.  Effectively, this embeds your application in IPython
-  rather than the other way round.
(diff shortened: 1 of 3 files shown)


takluyver:
Ping @ccordoba12 @wmvanvliet :-)

wmvanvliet:
I'm currently runnign python2.7 + PyQt4. This example sort of works for me (after porting it to PyQt4), but throws errors all over the place, complaining that some strings are `QString` instead of python strings and thus miss all kinds of methods. I'll try later on a different machine with python3.6 + PyQt5.

takluyver:
For PyQt4, you probably need `sip.setapi('QString', 2)`. But I'd argue that it's fine for example code to be written with the latest and greatest (i.e. Python 3 and PyQt5).

wmvanvliet:
The kernel shutdown hook doesn't seem to be working (I have to kill the program to exit). But otherwise, this runs fine on python3.6 + PyQT5

takluyver:
The shutdown hook appears to work for me. Maybe there's a better signal to connect it to? I just guessed one that seemed to do what I wanted.

takluyver:
I'm going to merge this, because I'm confident it's an improvement as is. If we can find a better way to handle shutting down the kernel, we can change that later.

ccordoba12:
Thanks Thomas!

PR #250, opened 2017-11-29
Implement Block de/indent 

### LINKED ISSUE #249: Implementing Block indent in qtconsole
(Was https://github.com/ipython/ipython/issues/10927)

I'd like to see block indent implemented in qtconsole, it's a pet peeve I hit every day.
Wrapping a piece of code in an if tediously involves manually indenting N lines. Pasting in an indented
position also doesn't preserve the indentation of the block pasted.

If one of the developers would kindly point me in the right direction, I'll like to code this up. And If there's interest, a PR may follow.

JeremySikes73:
Here's the first cut for #249. Before merging I'd like to use it for a week or so to flush out any kinks I may have introduced, but code review and testing by other interested folks is welcome.

Intention is to:
- Preserve current indent behavior when no selection is active.
- If selection **is** active, de/indent whole block by same amount, as determined by the column of the first line in the selection (indent is still done to next 4-space boundary).

Please advise on use of existing `self._tab_width=8` vs. (the new) `self._indent_width=4`.

Also noticed the criteria for triggering completion is puzzling: https://github.com/jupyter/qtconsole/blob/8508ce1e815867743dd4bb84534ebe4cfebbdc11/qtconsole/frontend_widget.py#L279

That's *"before"* and not *"right before"* , which means that:
```
In [1]:  Foo        |
```
Hitting TAB at the "|" cursor position really does trigger completion. Odd behavior.

closes #249

### DIFF
--- a/qtconsole/console_widget.py
+++ b/qtconsole/console_widget.py
@@ -906,7 +906,7 @@ class ConsoleWidget(MetaQObjectHasTraits('NewBase', (LoggingConfigurable, superQ
         """ Called when the tab key is pressed. Returns whether to continue
             processing the event.
         """
-        return False
+        return True
 
     #--------------------------------------------------------------------------
     # 'ConsoleWidget' protected interface
--- a/qtconsole/console_widget.py
+++ b/qtconsole/console_widget.py
@@ -1731,11 +1731,18 @@ class ConsoleWidget(MetaQObjectHasTraits('NewBase', (LoggingConfigurable, superQ
     def _get_leading_spaces(self):
         """ Get the number of leading spaces of the current line.
         """
-        cur = self._get_cursor()
-        cur.select(QtGui.QTextCursor.LineUnderCursor)
-        text = cur.selectedText()[len(self._continuation_prompt):]
-        return len(text) - len(text.lstrip())
 
+        cursor = self._get_cursor()
+        start_line = cursor.blockNumber()
+        if start_line == self._get_prompt_cursor().blockNumber():
+            # first line
+            offset = len(self._prompt)
+        else:
+            # continuation
+            offset = len(self._continuation_prompt)
+        cursor.select(QtGui.QTextCursor.LineUnderCursor)
+        text = cursor.selectedText()[offset:]
+        return len(text) - len(text.lstrip())
 
     @property
     def _prompt_pos(self):
--- a/qtconsole/console_widget.py
+++ b/qtconsole/console_widget.py
@@ -275,7 +275,7 @@ class ConsoleWidget(MetaQObjectHasTraits('NewBase', (LoggingConfigurable, superQ
         self._prompt_sep = ''
         self._reading = False
         self._reading_callback = None
-        self._tab_width = 8
+        self._tab_width = 4
 
         # List of strings pending to be appended as plain text in the widget.
         # The text is not immediately inserted when available to not
--- a/qtconsole/console_widget.py
+++ b/qtconsole/console_widget.py
@@ -1343,24 +1343,11 @@ class ConsoleWidget(MetaQObjectHasTraits('NewBase', (LoggingConfigurable, superQ
             elif key == QtCore.Qt.Key_Tab:
                 if not self._reading:
                     if self._tab_pressed():
-                        # real tab-key, insert four spaces
-                        step = 4 - (self._get_cursor().columnNumber() -
-                                    len(self._continuation_prompt)) % 4
-                        cursor.insertText(' '*step)
+                        self._indent(dedent=False)
                     intercepted = True
 
             elif key == QtCore.Qt.Key_Backtab:
-                cur = self._get_cursor()
-                cur.movePosition(QtGui.QTextCursor.StartOfLine)
-                cur.movePosition(QtGui.QTextCursor.Right,
-                                 QtGui.QTextCursor.MoveAnchor,
-                                 self._get_prompt_cursor().columnNumber())
-                spaces = self._get_leading_spaces()
-                step = spaces % 4
-                cur.movePosition(QtGui.QTextCursor.Right,
-                                 QtGui.QTextCursor.KeepAnchor,
-                                 min(spaces, step if step != 0 else 4))
-                cur.removeSelectedText()
+                self._indent(dedent=True)
                 intercepted = True
 
             elif key == QtCore.Qt.Key_Left:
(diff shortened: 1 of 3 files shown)


qtconsole/console_widget.py
```diff
@@ -898,7 +898,7 @@ def _tab_pressed(self):
         """ Called when the tab key is pressed. Returns whether to continue
             processing the event.
         """
-        return False
+        return True
```
JeremySikes73: Not sure what the larger implications are of this change, but without it, the event handling code for tab is untestable, because it is guarded by [an if](https://github.com/JeremySikes73/qtconsole/blob/2a903ef6988a2da8f5cb09cb881915e9effd245e/qtconsole/console_widget.py#L1337)

JeremySikes73:
I've been using this for a couple of days and so far i've hit no regressions or bad behaviour. Now that it has a few tests I think it's good to go. @ccordoba12, what do you think?

ccordoba12:
I'd prefer @wmvanvliet to give it look before merging. He has worked more with this part of the code.

wmvanvliet:
> Please advise on use of existing self._tab_width=8 vs. (the new) self._indent_width=4.

I would prefer if we didn't need to make a distinction between tab-width and indent-width. Looking at the existing code, `_tab_width` is initialized as 8 in `ConsoleWidget`, but in `FrontendWidget` redefined to be 4. Looks like we can just change `_tab_width` to be 4 everywhere, since that is the value we're actually using.

takluyver:
At some point it would be good to collapse the many layers of inheriting widget classes - it makes the code very confusing to work with, and I don't think anyone actually uses the lower layers independently of the upper ones.

It needs to be done carefully, though - among other things, the class names are used for setting config.

wmvanvliet:
I get an error:

```
[IPKernelApp] ERROR | Exception in message handler:block_indent] 
  File "/work/modules/Ubuntu/14.04/amd64/common/anaconda3/latest/envs/neuroimaging/lib/python3.6/site-packages/ipykernel/kernelbase.py", line 235, in dispatch_shell
    handler(stream, idents, msg)
  File "/work/modules/Ubuntu/14.04/amd64/common/anaconda3/latest/envs/neuroimaging/lib/python3.6/site-packages/ipykernel/kernelbase.py", line 434, in complete_request
    matches = self.do_complete(code, cursor_pos)
  File "/work/modules/Ubuntu/14.04/amd64/common/anaconda3/latest/envs/neuroimaging/lib/python3.6/site-packages/ipykernel/ipkernel.py", line 257, in do_complete
    txt, matches = self.shell.complete('', line, line_cursor)
  File "/work/modules/Ubuntu/14.04/amd64/common/anaconda3/latest/envs/neuroimaging/lib/python3.6/site-packages/IPython/core/interactiveshell.py", line 1996, in complete
    return self.Completer.complete(text, line, cursor_pos)
  File "/work/modules/Ubuntu/14.04/amd64/common/anaconda3/latest/envs/neuroimaging/lib/python3.6/site-packages/IPython/core/completer.py", line 1925, in complete
    return self._complete(line_buffer=line_buffer, cursor_pos=cursor_pos, text=text, cursor_line=0)[:2]
  File "/work/modules/Ubuntu/14.04/amd64/common/anaconda3/latest/envs/neuroimaging/lib/python3.6/site-packages/IPython/core/completer.py", line 1998, in _complete
    cursor_pos, cursor_line, full_text)
  File "/work/modules/Ubuntu/14.04/amd64/common/anaconda3/latest/envs/neuroimaging/lib/python3.6/site-packages/IPython/core/completer.py", line 1348, in _jedi_matches
    text, namespaces, column=cursor_column, line=cursor_line + 1)
  File "/work/modules/Ubuntu/14.04/amd64/common/anaconda3/latest/envs/neuroimaging/lib/python3.6/site-packages/jedi/api/__init__.py", line 386, in __init__
    super(Interpreter, self).__init__(source, **kwds)
  File "/work/modules/Ubuntu/14.04/amd64/common/anaconda3/latest/envs/neuroimaging/lib/python3.6/site-packages/jedi/api/__init__.py", line 102, in __init__
    raise ValueError('`column` parameter is not in a valid range.')
ValueError: `column` parameter is not in a valid range.
``

When placing the cursor on the first column by pressing the HOME key, then press SHIFT-END to select to the last column (including a newline I suppose) and then pressing TAB.

**Comment by wmvanvliet** [comment]
Other than that, I'm loving this new block indent feature!

File: qtconsole/console_widget.py [hunk 2]
```diff
@@ -1722,9 +1716,17 @@ def _get_last_lines_from_list(self, text_list, num_lines):
     def _get_leading_spaces(self):
         """ Get the number of leading spaces of the current line.
         """
-        cur = self._get_cursor()
-        cur.select(QtGui.QTextCursor.LineUnderCursor)
-        text = cur.selectedText()[len(self._continuation_prompt):]
+
+        cursor = self._control.textCursor()
```
wmvanvliet: `cursor = self._get_cursor()`

I know it's a bit silly, but we use `_get_cursor()` everywhere else. No good reason to change it here.

qtconsole/console_widget.py
```diff
@@ -1872,6 +1874,44 @@ def _get_word_end_cursor(self, position):
...
 
+    def _indent(self, dedent=True):
+        """ Indent/Dedent current line or current text selection.
+        """
+        num_newlines = self._get_num_lines_selection()
+        save_cur = self._get_cursor()
+        cur = self._get_cursor()
+
+        # - move to first line of selection, if present
+        # - calculate number of space to indent it to 4-space multiple
```
wmvanvliet: Why have a "list" of two items here? Just place the comment above the relevant line in the code.

qtconsole/console_widget.py
```diff
@@ -1872,6 +1874,44 @@ def _get_word_end_cursor(self, position):
...
+        cur.setPosition(cur.selectionStart())
+        self._control.setTextCursor(cur)
+        spaces = self._get_leading_spaces()
```
wmvanvliet: # Calculate number of space to indent it to 4-space multiple
    spaces = self._get_leading_spaces()

qtconsole/console_widget.py
JeremySikes73: Changed here and in the tests

qtconsole/console_widget.py
JeremySikes73: Fixed

qtconsole/console_widget.py
JeremySikes73: Fixed

JeremySikes73:
@wmvanvliet, I can't reproduce https://github.com/jupyter/qtconsole/pull/250#issuecomment-348927449

What I get is a kitchen sink completion. The backtrace also doesn't show anything related to the code changed. Can you provide more specific steps to reproduce?

I'm using jedi 0.11.

qtconsole/console_widget.py
JeremySikes73: @wmvanvliet , no problem here?

wmvanvliet:
I'm on commit `11464fc8392e00369a609880a418361e212e42d7`, using jedi 0.11.0.

Steps to reproduce:

`$ jupyter qtconsole &`

Enter the following, but don't execute the cell:

```
"foo"
"bar"
```
Place cursor at the beginning of the second line (`"bar"`). Press SHIFT+END to select the whole line. Press TAB. 

A kitchen sink completion is also not the desired behaviour, right?

JeremySikes73:
I still see no traceback or error, but I do see the completion issue. I've added a commit to address this.

JeremySikes73:
The commit history has gotten too crowded with false starts, I've rewitten it to make the changes clearer.

I've also removed the changes @wmvanvliet requested in https://github.com/jupyter/qtconsole/pull/250#discussion_r154616312 because the "offending" code was actually conformant to preexisting code. We can make those changes in a separate PR, if needed, and do it comprehensively.

wmvanvliet:
Bug is still there. Also, having a single line of input like this:

```
In [1]: "foo"
```

And selecting the whole line with HOME, SHIFT+END and then pressing TAB triggers kitchen sink completion.

JeremySikes73:
@wmvanvliet, sorry but I just can't reproduce this (using yesterday's 7d7b861). 

Following your instructions, I end up with:

foo
as expected.

wmvanvliet:
Interesting. What platform are you on? (QT has different default keybindings depending on the platform) I'm currently on Linux.

JeremySikes73:
me too.

any (custom input transformers](http://ipython.readthedocs.io/en/stable/config/inputtransforms.html) in your ipython startup files?

wmvanvliet:
Nope. I'll debug a little on my end.

wmvanvliet:
Ah, my mistake. I wasn't on the latest commit due to rebasing issues. The change to `complete = non_ws_before and self._get_cursor().selectedText() == ''` did the trick. Now it works!

wmvanvliet:
It also resolves the jedi error I was getting.

wmvanvliet:
Ok, this LGTM. +1 To merge this from my end. Thanks @JeremySikes73!

JeremySikes73:
:birthday:

ccordoba12:
Thanks @JeremySikes73! Great addition!

PR #251, opened 2017-11-30
Fix paste dedent corner case

JeremySikes73:
Starting with the following snippet
```
If True:       
    if False:        # line A
        pass  # line B
```
If I copy Block AB, selecting two complete lines, with indentation, and paste it into qtconsole in an indented position:
```
While Foo:
    _ # cursor here, <paste>
```
The result is:
```
While Foo:
    if False:        # line A
    pass  # line B
```
When I'd expect to see:
```
While Foo:
    if False:        # line A
        pass  # line B
```

### DIFF
--- a/qtconsole/console_widget.py
+++ b/qtconsole/console_widget.py
@@ -68,7 +68,7 @@ class ConsoleWidget(MetaQObjectHasTraits('NewBase', (LoggingConfigurable, superQ
     )
     execute_on_complete_input = Bool(True, config=True,
         help="""Whether to automatically execute on syntactically complete input.
-        
+
         If False, Shift-Enter is required to submit each execution.
         Disabling this is mainly useful for non-Python kernels,
         where the completion check would be wrong.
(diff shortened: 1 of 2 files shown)


qtconsole/console_widget.py
```diff
@@ -768,6 +768,15 @@ def paste(self, mode=QtGui.QClipboard.Clipboard):
             # Remove any trailing newline, which confuses the GUI and forces the
             # user to backspace.
             text = QtGui.QApplication.clipboard().text(mode).rstrip()
+
+            # dedent removes "common leading whitespace" but to preserve relative
+            # indent of multiline code, we have to compensate for any
+            # leading space on the first line, if we're pasting into
+            # an indented position.
+            cursor_offset = cursor.position()-self._get_line_start_pos()
```
ccordoba12: Please add spaces around `-`

JeremySikes73: Done

qtconsole/console_widget.py
```diff
@@ -768,6 +768,15 @@ def paste(self, mode=QtGui.QClipboard.Clipboard):
             # Remove any trailing newline, which confuses the GUI and forces the
             # user to backspace.
             text = QtGui.QApplication.clipboard().text(mode).rstrip()
+
+            # dedent removes "common leading whitespace" but to preserve relative
+            # indent of multiline code, we have to compensate for any
+            # leading space on the first line, if we're pasting into
+            # an indented position.
+            cursor_offset = cursor.position() - self._get_line_start_pos()
+            if text.startswith(' '*cursor_offset):
```
ccordoba12: Also here, spaces around `*`

PR #228, opened 2017-07-26
Add unmatched bracket highlighting capability

jnsebgosselin:
This PR is to add the capability to highlight unmatched bracket in red in the console.

ipython_unmatched

PR #253, opened 2017-12-01
Colorize unmatched brackets and improve matched bracket handling

### LINKED ISSUE #252: Quirk in BracketMatcher
quirk1

JeremySikes73:
#228 makes a start on a great feature to have, but doesn't quite go far enough to get there.

Unmatched brackets should be highlighted regardless of the position of the cursor.
Along the way I found that the *matched* bracket identification was too naive, so this fixes #252 as well.

The first commit makes the `ConsoleWidget` an argument of BracketMatcher. Strictly speaking, this is
a layering violation, it would be nicer if BracketMatcher didn't know anything about it. Unfortunately, since the underlying textEdit mixes code and prompts, and output, and multiple cells, which means it's just not reasonable to do syntax-dependent things on a raw buffer. 

Money shot:

money

In #228 there was a code review comment about customizing colors from CSS. Show me an example and I'll implement it. The existing highlighting of matched brackets doesn't do it either.

Lots of tests included.

closes #252
closes #228 (supersedes, actually)

### DIFF
--- a/qtconsole/rich_jupyter_widget.py
+++ b/qtconsole/rich_jupyter_widget.py
@@ -125,6 +125,8 @@ class RichJupyterWidget(RichIPythonWidget):
             prompt_number = content.get('execution_count', 0)
             data = content['data']
             metadata = msg['content']['metadata']
+            if not self.from_here(msg):
+                self._append_plain_text('\n', True)
             if 'image/svg+xml' in data:
                 self._pre_image_append(msg, prompt_number)
                 self._append_svg(data['image/svg+xml'], True)


jnsebgosselin:
Thank you for taking this over! My PR was not going anywhere...

> Unmatched brackets should be highlighted regardless of the position of the cursor.
I'm not sure about this, but this may be a good idea.  I'll test it as soon as I can.

Also, running this in Spyder gives me this error:
```python
  File "C:\Users\jsgosselin\spyder\spyder\plugins\base.py", line 175, in <lambda>
    toggled=lambda checked: self.toggle_view(checked),
  File "C:\Users\jsgosselin\spyder\spyder\plugins\ipythonconsole.py", line 711, in toggle_view
    self.create_new_client(give_focus=False)
  File "C:\Users\jsgosselin\spyder\spyder\plugins\ipythonconsole.py", line 982, in create_new_client
    options_button=self.options_button)
  File "C:\Users\jsgosselin\spyder\spyder\widgets\ipythonconsole\client.py", line 125, in __init__
    local_kernel=True)
  File "C:\Users\jsgosselin\spyder\spyder\widgets\ipythonconsole\shell.py", line 82, in __init__
    self.set_bracket_matcher_color_scheme(self.syntax_style)
  File "C:\Users\jsgosselin\spyder\spyder\widgets\ipythonconsole\shell.py", line 133, in set_bracket_matcher_color_scheme
    self._bracket_matcher.format.setBackground(mpcolor)
AttributeError: 'BracketMatcher' object has no attribute 'format'
```

The problem is introduced by the changes made [here](https://github.com/jupyter/qtconsole/pull/253/files#diff-6720cdbfcdacabd429e63a46d431a79eR103). That would be great if backward compatibility could be preserved while implementing this feature.

goanpeca:
>Unmatched brackets should be highlighted regardless of the position of the cursor.

Hmmm could you post a few more examples of this looks, just to get the feel of it? maybe on multilines as well?

JeremySikes73:
@jnsebgosselin , Certainly back-compatibility should be maintained when possible. I've added back the removed attribute. But note that this PR added an additional format/style, so if Spyder wants to adjust that as well, you'll have to switch your code to use the `formats` dictionary. 

Here are more screenshots (note the cursor/position is visible in all screenshots above and below):


after

The differences are:
- Unmatched brackets are highlighted in red as long as they are unmatched, or until the cell is executed, at which time they are cleared. This is a syntax error and so highlighting does not depend on the cursor position as in #228.
- Matched brackets are still highlighted and still only when the cursor is right next to them, but only outside quoted strings. This is a byproduct of fixing the old and overly naive behaviour of: 
quirk

goanpeca:
Thanks @jnsebgosselin for checking this. @JeremySikes73 thanks, it is now clear :-) and definitely and improvement. @ccordoba12 any other thoughts?

jnsebgosselin:
> or until the cell is executed, at which time they are cleared. 

That is what I was concerned about. Good job. 👍 I also think this is a nice improvement.

JeremySikes73:
Thanks. 

> That is what I was concerned about. 

for the record:

residual

JeremySikes73:
Any more feedback from committers? going on two weeks here...

jnsebgosselin:
@ccordoba12  Sorry for the delay in testing this.

I've tested it on Python 3.6.4 64bits, Qt 5.9.2, PyQt5 5.9.1 on Windows  7 for Spyder 4.0.0.dev0 and Spyder 3.2.7.dev0 (Commit: d85dcbd19) and there does not seem to be any compatibility problem anymore.

@JeremySikes73 This is a great new addition and I like it. However, is the result in the screenshot below the expected behaviour? I'm under the impression that there is too much unmatched bracket highlighting now.

image

JeremySikes73:
@jnsebgosselin , your screenshot shows the expected behaviour.  Opening/closing brackets are always paired off from the inside outwards (as in a stack). In this case we get the pairs `(}`,`{]` and `[` (unpaired), so every bracket is unmatched and therefore every bracket is highlighted.

jnsebgosselin:
I understand, thanks for your explanation. But from a user perspective, I think that would be more useful if  `[]` and `{}` could be matched correctly, so that the highlighting pinpoints exactly where the problem is. This is also the behavior in the Spyder editor. So I think that would be preferable if the matching/unmatching behavior was the same in the console and the editor in Spyder. @ccordoba12  what do you think?

brackets_unmatched

Also, if I add a `)` bracket at the end of my example, the `()` are matched correctly, but the `{}` and `[]` remains unmatched.

image

But, if I use `)` instead of `(`, the `{}` and `[]` are matched correctly then.

image

ccordoba12:
> what do you think?

I also agree. Showing all braces as unmatched is really unhelpful.

jnsebgosselin:
@JeremySikes73 We are sorry that you've decided to close your PR. I'm going to try continuing the good work that you've started in another PR, so that your contribution do not go to waste.

PR #254, opened 2017-12-02
DONOTMERGE: Further Enhancements to indentation handling on paste

JeremySikes73:
This is a followup to #251 

This change is required to handle another corner case I've found. 

All in all, here are the case I suggest should be handled:
i1
i2
i3
i5

When (general case) pasted in this position:
it

#251 falls short. 

With this change this case:
i4
also becomes pretty when pasted. But that's just a side-effect of actually handling the corner case mentioned.

**TODO**: 
- [X] Add tests

### DIFF
--- a/qtconsole/rich_jupyter_widget.py
+++ b/qtconsole/rich_jupyter_widget.py
@@ -125,6 +125,8 @@ class RichJupyterWidget(RichIPythonWidget):
             prompt_number = content.get('execution_count', 0)
             data = content['data']
             metadata = msg['content']['metadata']
+            if not self.from_here(msg):
+                self._append_plain_text('\n', True)
             if 'image/svg+xml' in data:
                 self._pre_image_append(msg, prompt_number)
                 self._append_svg(data['image/svg+xml'], True)


qtconsole/console_widget.py
```diff
@@ -760,6 +760,9 @@ def paste(self, mode=QtGui.QClipboard.Clipboard):
             used to access the selection clipboard in X11 and the Find buffer
             in Mac OS. By default, the regular clipboard is used.
         """
+        def num_leading(s):
+            return len(s)-len(s.lstrip())
```
ccordoba12: Please add spaces around `-`

qtconsole/console_widget.py
```diff
@@ -769,15 +772,41 @@ def paste(self, mode=QtGui.QClipboard.Clipboard):
...
+            # remove common leading space, and leading space on first line
+            text = dedent(text.lstrip())
+
+            lines=text.splitlines()
+            # Quantize all relative indents to multiple of 4
+            for i, l in enumerate(lines):
+                cur_n = num_leading(lines[i])
+                if cur_n % 4 > 0:
+                    cur_n += 4 - cur_n % 4
+                    lines[i] = (' ' *((1+cur_n//4)*4)) + lines[i].strip()
```
ccordoba12: Spaces around `*`

qtconsole/console_widget.py
```diff
@@ -769,15 +772,41 @@ def paste(self, mode=QtGui.QClipboard.Clipboard):
...
+            # the result of the user copying a snippet without the
+            # first lines' indentation. If we don't fix it, we'll
+            # actually *introduce* the problem. So we quantize the indent.
+            new_lines = [lines[0]] + ([''] * len(lines))
+            for i, (prev, cur) in enumerate(zip(lines[:-1],lines[1:]), 1):
+                prev_n, cur_n = num_leading(prev), num_leading(cur)
+                new_prev_n = num_leading(new_lines[i-1])
+                new_line = cur
+                if cur_n - prev_n > 4:
+                    new_line = (' '*(new_prev_n+4)) + cur.lstrip()
```
ccordoba12: Spaces around `*`

ccordoba12:
Thanks for this @JeremySikes73! It seems you're assuming all code is formatted with 4 spaces, but what would happen if the user is using a different level of indentation, like 2 spaces?

JeremySikes73:
Well, IPython is currentyl hardwired to indent by 4-spaces on tab. So that's the natural tab-width when working in IPython by my logic. I haven't gotten an answer in #250 about `_tab_width` and `_indent_width`. Once you clear that up, and merge, I can have this use the same setting. But it doesn't matter anyway unless the global setting is configurable.

takluyver:
I'm not following this PR closely, but don't forget that the Qt console can be used with kernels other than IPython, and languages other than Python. So don't build the indentation behaviour solely around Python conventions!

JeremySikes73:
I can't think of a language where normalizing the width of indentation across lines would disrupt anything. Do you have a concrete counter-example we can discuss?

If different languages are supported, you must have implemented a mechanism for introducing language-specific behavior. Show me what it is and I'll gladly make any syntax-specific changes apply to python kernels only.

takluyver:
I don't think it would break anything, but being zealous about four-space indentation may not be to the taste of every language community. I'm not saying it's a bad idea - I haven't really thought about this - I just wanted to make sure we weren't designing something too Python-specific.

The automatic indentation when you're typing code communicates with the kernel. When you press enter, it sends an [is_complete_request](http://jupyter-client.readthedocs.io/en/latest/messaging.html#code-completeness). The response can indicate the indentation for the next line. There's no specific messaging for re-indenting a block of code. It could make an `is_complete_request` for each line, but that's probably a bad idea when pasting a large block of code.

JeremySikes73:
(Unrelated  to talkuyver's comments)

I should be probably be clearer about what this is trying to fix, if only for my own benefit.

The underlying issue is that the following two cases yield identical clipboard contents
i9
i10

Since we can't know what the original was, we have which version you coerce to when pasting. 
And, in this case, you can't really say which one is more correct.
But, that choice also determines the result in the following case (and nth case):
i11

Before this PR:
i12

After:
i13

So I think this new behaviour makes sense. Of course if the user is responsible enough to include the leading space on all lines in the copied selection, there's no problem either way.

Tests are pending.

JeremySikes73:
@takluyver, I've already offered to make it python-specific if it's technically feasible.

I'm not sure I see how the knowledge of where the next line should indent can be used to determine the prefered indent width. In any case, I have no special preference for 4-space indent (well, I do, but only personally). There's simply no configuration setting I can take the preferred value from. I've mentioned that above and in #250.

More to the point, I do understand your concerns (and would hate to break other people's UX - I can do without the hate thanks). But, on the one hand this is the vague kind of uncertainty which stops everything cold, and yet without concrete examples I really have no way to address the problem, assuming it exists. What would you have me do?

takluyver:
Feel free to carry on doing what you are doing - it doesn't sound like this needs to be Python specific, and if it causes problems we can always make more changes (or revert it if there's really no way to fix it). I mostly just wanted to make sure you're aware that the code being pasted may not be Python.

However, looking at your code, what happens if I paste something indented with two spaces, like:

```python
for a in range(5):
  if a % 2 == 0:
    print('odd')
```

It looks like this is going to become:

```python
for a in range(5):
    if a % 2 == 0:
    print('odd')
```

Which would be invalid. Two spaces is not uncommon, e.g. in Google projects like [Chromium OS](https://www.chromium.org/chromium-os/python-style-guidelines).

JeremySikes73:
Good point. Irealize @ccordoba12 asked about this earlier but I didn't realize the implications.

But I think I can address this if I just normalize indents more carefully, I'll give it a try.

JeremySikes73:
I've changed the way quantization is done. Now it just checks if a line is indented (at all) relative to its parent, and if it is sets the relative indent to `self._tab_width`. 

It's really time to write those tests, but afaict this passes all the cases mentioned so far, including the most recent 2-space indents case.

JeremySikes73:
Another case (currently failing):
i15

JeremySikes73:
New revision. 

The code now deals properly with whatever the original indent width was used, and it ses `ConsoleWidget._tab_width` as its own width parameter.

I've now **added tests** for all the cases discusses I've taken extra special care to make the tests as readable and self-documenting as possible. 

The tests explain the reasons why choices were made, with examples. 
There are explicit cases (case, really) for *unfortunate* behaviour  that show the problem, show why it occurs, and show how to avoid it. 

Overall, the principle for a joyous life is quite simple: **always include the leading space when you copy**, then everything should work. When the joyous life principle is violated things only work *mostly perfectly*.

I can add a small section to the documentation about the new "paste smartiness", If you think it necessary.

takluyver:
I think it would be good to have a note in the changelog, thanks!

JeremySikes73:
Added. Can we tag this 4.4?

JeremySikes73:
When pasting with an active selection, the paste insert position != cursor position. Added test case and code to handle it.

case

JeremySikes73:
This is probably as solid as I can get it without at this point. Removed WIP status.

JeremySikes73:
Rebased

JeremySikes73:
Still hitting new failure cases. I'll keep pushing improvements for my own use, but the fact that despite my efforts and lots of tests really nasty surprises still crop up from time time, this feels too risky to merge.

PR #255, opened 2017-12-02
DONOTMERGE: Copy with no selection copies current cell contents

JeremySikes73:


JeremySikes73:
Thanks. I really wanted to have the ubiquitous C-a be "select all" instead but that was taken, so this is a compromise.

As for whitespace, there seems to be no way in your guidelines to deal with this, except to make a change to every function in every file. In addition, you don't flag it in code review when new dead whitespace is introduced (#228). It's an inconvenience to contributors whose editors do the right thing and strips trailing whitespace. It gets pretty tedious to spent time each PR putting back the noise in the file.

Anyway I get the impression that development on qtconsole is quite leisurely these days, so the conflict argument doesn't really have as much weight as it would have in a very active project.

How about giving this a pass?

wmvanvliet:
How about just removing all trailing whitespace in a separate PR? #256 .

JeremySikes73:
@wmvanvliet, that would even better. I'll rebase once that's merged.

JeremySikes73:
I think this harmless hidden "hack" becomes unnecessary if #258 is merged, so hold off.

JeremySikes73:
#258 had been merged, making this hack unnecessary.

PR #256, opened 2017-12-04
Eliminate training whitespace

wmvanvliet:
Get rid of all training whitespace. This gets placed by accident. Let's clean it up!

Ran the following on the source code:

    python -m autopep8 --select W2 -i qtconsole/*.py
    python -m autopep8 --select W2 -i qtconsole/tests/*.py

### DIFF
--- a/qtconsole/base_frontend_mixin.py
+++ b/qtconsole/base_frontend_mixin.py
@@ -136,12 +136,12 @@ class BaseFrontendMixin(object):
         handler = getattr(self, '_handle_' + msg_type, None)
         if handler:
             handler(msg)
-    
+
     def from_here(self, msg):
         """Return whether a message is from this session"""
         session_id = self._kernel_client.session.session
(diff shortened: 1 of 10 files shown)


ccordoba12:
@takluyver, what do you think about this one?

takluyver:
I generally think it's not worth it, but this is a fairly small diff, and it avoids having to ask people to redo changes where they've accidentally stripped out the whitespace. You and @wmvanvliet are doing most of the maintenance on qtconsole now, so it's up to you!

wmvanvliet:
It's a small thing yes, but it took me only 5 seconds to run the command to auto-trim whitespace everywhere, it was either that or taking another sip of coffee :)

takluyver:
I mean that because it's small, it's unlikely to cause much pain by conflicting with other changes. :-)

Let's do it.

CAM-Gerlach:
Just FYI—its "trailing" whitespace, not "training".

PR #260, opened 2017-12-13
Fix behaviour of right arrow key

### LINKED ISSUE #245: Rightarrow at the end of a line of a multiline input sends cursor all the way to end of input
## Description

In `qtconsole`, either launched standalone or within Spyder, if I paste (or type) a multiline statement into the console, put my cursor at the end of any line, and press rightarrow, instead of going to the first char on the next line as one would expect, the cursor will instead jump all the way to the end of the last line of the statement (as if I pressed Ctrl-End). 

There seems to be no logical reason why this would be a feature, as one could simply press the aforementioned shortcut for the same effect, and it leaves no easy way to go to the first char on the next line other then downarrow + home; moreover, it breaks basic user expectations on how such text-based interfaces generally behave. As @wmvanvliet helpfully suggested it was probably a bug, I'm posting it here. I've tested this on multiple m,achines, OS versions and spyder/qtconsole installs, including a fresh anaconda install and the latest dev version as of two weeks ago, and all experienced the same behavior. As far as I'm aware, this has been the case for as long as I've been using qtconsole; only a couple months now.

**What steps will reproduce the problem?**

1. Copy or type a multiline statement into the console, e.g.
```
print("foo" + "bar"
      + "This is a "|
      + "test string "
      + "for qtconsole.")
```
2. Place cursor where indicated with "|" (or at the end of any other line)
3. Press rightarrow once

**What is the expected output? What do you see instead?**
For the above example, Expected:
```
print("foo" + "bar"
      + "This is a "
     |+ "test string "
      + "for qtconsole.")
```
Actual:
```
print("foo" + "bar"
      + "This is a "
      + "test string "
      + "for qtconsole.")|
```

## Version and main components

* qtconsole 4.3.1 and 4.4.0-dev (as of 2 weeks ago)
* IPython 6.1.0
* Spyder Version: 3.2.4 
* Python Version: 3.6.3
* Qt Versions: 5.6.2, PyQt5 5.6 on Windows 8.1 Pro x64 and Windows 10 Ent x64

sys_info():
```
{'commit_hash': 'd86648c5d',
 'commit_source': 'installation',
 'default_encoding': 'utf-8',
 'ipython_path': 'C:\\Anaconda3\\lib\\site-packages\\IPython',
 'ipython_version': '6.1.0',
 'os_name': 'nt',
 'platform': 'Windows-8.1-6.3.9600-SP0',
 'sys_executable': 'C:\\Anaconda3\\python.exe',
 'sys_platform': 'win32',
 'sys_version': '3.6.3 |Anaconda, Inc.| (default, Oct 15 2017, 03:27:45) [MSCv.1900 64 bit (AMD64)]'}
```

## Spyder/Jupyter Dependencies
```
pyflakes >=0.6.0 :  1.6.0 (OK)
pycodestyle >=2.3:  2.3.1 (OK)
pygments >=2.0   :  2.2.0 (OK)
pandas >=0.13.1  :  0.20.3 (OK)
numpy >=1.7      :  1.13.3 (OK)
sphinx >=0.6.6   :  1.6.3 (OK)
rope >=0.9.4     :  0.10.5 (OK)
jedi >=0.9.0     :  0.10.2 (OK)
nbconvert >=4.0  :  5.3.1 (OK)
sympy >=0.7.3    :  1.1.1 (OK)
cython >=0.21    :  0.26.1 (OK)
qtconsole >=4.2.0:  4.3.1 (OK)
IPython >=4.0    :  6.1.0 (OK)
pylint >=0.25    :  1.7.4 (OK)

```

wmvanvliet:
Currently, pressing the right arrow key when at the end of a line always moves the cursor to the end of the cell, even when the cell has multiple lines. The intended behaviour was to move the cursor to the next line (respecting the continuation prompt). See #245 for more details.

Fixes #245

### DIFF
--- a/qtconsole/console_widget.py
+++ b/qtconsole/console_widget.py
@@ -1367,13 +1367,16 @@ class ConsoleWidget(MetaQObjectHasTraits('NewBase', (LoggingConfigurable, superQ
                     intercepted = not self._in_buffer(position - 1)
 
             elif key == QtCore.Qt.Key_Right:
-                original_block_number = cursor.blockNumber()
-                self._control.moveCursor(QtGui.QTextCursor.Right,
-                                mode=anchormode)
-                if cursor.blockNumber() != original_block_number:
+                #original_block_number = cursor.blockNumber()
+                if position == self._get_line_end_pos():
+                    cursor.movePosition(QtGui.QTextCursor.NextBlock, mode=anchormode)
+                    cursor.movePosition(QtGui.QTextCursor.Right,
+                                        mode=anchormode,
+                                        n=len(self._continuation_prompt))
+                    self._control.setTextCursor(cursor)
+                else:
                     self._control.moveCursor(QtGui.QTextCursor.Right,
-                                        n=len(self._continuation_prompt),
-                                        mode=anchormode)
+                                             mode=anchormode)
                 intercepted = True
 
             elif key == QtCore.Qt.Key_Home:
--- a/qtconsole/tests/test_console_widget.py
+++ b/qtconsole/tests/test_console_widget.py
@@ -246,6 +246,21 @@ class TestConsoleWidget(unittest.TestCase):
         self.assertEqual(w._get_input_buffer(),
                          ("foo = ['foo', 'foo', 'foo',    \n"
                           "'bar', 'bar', 'bar']"))
+
+        # Left and right keys should respect the continuation prompt
+        w._set_input_buffer("line 1\n"
+                            "line 2\n"
+                            "line 3")
+        c = control.textCursor()
+        c.setPosition(20)  # End of line 1
+        control.setTextCursor(c)
+        QTest.keyClick(control, QtCore.Qt.Key_Right)
+        # Cursor should have moved after the continuation prompt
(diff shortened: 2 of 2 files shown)


ccordoba12:
@CAM-Gerlach, please help us to test this one.

CAM-Gerlach:
@ccordoba12 Already on it!

CAM-Gerlach:
@wmvanvliet @ccordoba12 Works as requested. Thanks!

By the way, does `qtconsole` not have a requirements.txt or equivalent? I just figured out and installed the necessary dependencies manually (I can't remember what I did on my other dev machine where I had a qtconsole-dev environment set up already). Alternatively, I figure I could have just installed stable qtconsole into the environment with its dependencies, and then uninstalled the former without removing the latter, but that would be a bit of a hack. I'm still new to this, so I'm probably missing something...

ccordoba12:
Thanks @CAM-Gerlach!

> Alternatively, I figure I could have just installed stable qtconsole into the environment with its dependencies, and then uninstalled the former without removing the latter, but that would be a bit of a hack.

This is what I usually do with most packages, so I think it's fine.

PR #265, opened 2018-02-02
Add missing newline when printing remote results

mtikekar:
When qtconsole is mirroring another session, (e.g. `jupyter qtconsole --existing --JupyterWidget.include_other_output = True`) the output messages from the other session get printed on the same line as the input. This commit fixes that.

### DIFF
--- a/qtconsole/rich_jupyter_widget.py
+++ b/qtconsole/rich_jupyter_widget.py
@@ -125,6 +125,8 @@ class RichJupyterWidget(RichIPythonWidget):
             prompt_number = content.get('execution_count', 0)
             data = content['data']
             metadata = msg['content']['metadata']
+            if not self.from_here(msg):
+                self._append_plain_text('\n', True)
             if 'image/svg+xml' in data:
                 self._pre_image_append(msg, prompt_number)
                 self._append_svg(data['image/svg+xml'], True)


ccordoba12:
Thanks for your contribution. It'll be part of qtconsole 4.4, to be released shortly.

mtikekar:
I had made these changes on 4.3.1 and tested them there, but pushed them to master without testing them there. It turns out that the changes on master had already fixed the problem elsewhere (by adding the newline after the input, instead of adding it before the output as I did.)

It would be best to revert my commit.

ccordoba12:
@dalthviz, please open a new PR to revert this one.

takluyver:
There's a revert button by the line where you merged it, which will easily open a new PR. I don't know if the contributor has access to click that, or only the maintainers.

ccordoba12:
That creates a branch in the repo, that's why I don't like it. Besides, I think only maintainers have access to that.

mtikekar:
I do see the button, but yes, it will create a branch.

takluyver:
We can remove the branch again after merging. Go for it - there's no point doing it manually when Github makes it easy. :-)

dalthviz:
So are we going to use the revert button of Github or should I create the PR? Probably with Github this would take just a few clicks, right?

ccordoba12:
@dalthviz, I'll revert it. Let's see how this goes.

PR #280, opened 2018-04-16
Fix compatibility with PySide2 on Python 2.7

astrofrog:
This fixes a bug in commit_api which caused PySide2 binding to be forbidden even if the loaded API was PySide2 (I think it was just a typo in the if statement)

### DIFF
--- a/qtconsole/qt_loaders.py
+++ b/qtconsole/qt_loaders.py
@@ -74,7 +74,7 @@ def commit_api(api):
         ID.forbid('PySide')
         ID.forbid('PyQt4')
         ID.forbid('PyQt5')
-    if api == QT_API_PYSIDE:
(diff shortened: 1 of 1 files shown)


astrofrog:
For some reason this only manifests itself on Python 2.7 though, and it works fine on Python 3.6, which suggests maybe somehow the forbidding isn't working quite right on Python 3.6?

takluyver:
Yep, it certainly looks like it was just a typo. It could well be that the import denier stuff doesn't work right in Python 3.6 - it's hard to test. It's probably more important to fix the [copy in IPython](https://github.com/ipython/ipython/blob/master/IPython/external/qt_loaders.py), because that's what runs in the kernel, denying imports for user code.

PR #282, opened 2018-05-01
refresh situation on Qt Windows packages

stonebig:


ccordoba12:
Sorry, I don't agree with this. We don't want to start recommending alpha stuff to run qtconsole. Please reopen when PySide2 is officially released.

README.md
```diff
@@ -38,16 +38,21 @@ To install:
...
 ### Installing Qt (if needed)
 We recommend installing PyQt with [conda](http://conda.pydata.org/docs):
 
     conda install pyqt
 
 or with a system package manager. For Windows, PyQt binary packages may be
-used.
+used:
+
+    pip install pyqt5
```
takluyver: Actually, this now works (or should work) on 64-bit Linux and Mac as well.

README.md
```diff
@@ -38,16 +38,21 @@ To install:
...
+
+or
+
+    pip install http://download.qt.io/snapshots/ci/pyside/5.11/latest/pyside2/PySide2-5.11.0a1-5.11.0-cp36-cp36m-win_amd64.whl
```
takluyver: Does this work with `https`? We should never recommend installing software from an unsecure connection unless there's no other option (and if there's no other option, we should be trying to figure one out ASAP).

takluyver:
Oh, and I missed that it was an alpha. I'd agree with Carlos that it's not great to point people to alpha software.

ccordoba12:
PySide2 is going to be officially released with Qt 5.11, scheduled by the end of this month. I hope by then the Qt guys will upload PySide2 to PyPI, so there's no need to point people to an alpha wheel in an insecure location.

README.md
stonebig: I don't have a Mac nor a Linux VM, so I didn't dare writing about that. Maybe you can complete the update ?

stonebig:
https changed proof: 
https_pyside2

stonebig:
As I don't know what you can mean per secure, I removed nevertheless the https link.

README.md
takluyver: I think it does work, so I'd just remove the phrasing that says it's only for Windows.

takluyver:
Thanks - I think installing from a trusted site with https is OK, but we should leave it until PySide2 is actually released.

stonebig:
as a remark, shouldn't all the other link be changed to https ? even if documentation ?

takluyver:
It's certainly worth doing anywhere that that the https URL works. It's most critical for automated installation (IMO), because human readable information has to pass a sense check before it breaks things. But if you want to make PRs changing links to https, go for it!

takluyver:
Thanks! I assume you've checked that all those URLs work with https?

stonebig:
they all works, but the https://pyside.github.io/docs/pyside was reverting to http. until I put a "/" at the end to remain https

takluyver:
That sounds like a bug in Github's redirect rules, but thanks for checking :-)

README.md
ccordoba12: Thomas is right, this works for macOS and Linux as well.

ccordoba12:
Closing because this is really outdated.

PR #284, opened 2018-05-24
Check that the highlighting_on attribute exists before accessing it

ptosco:
I noticed that I occasionally get an `AttributeError` while using qtconsole:
```
  File "/build/Python/linux-x86_64/lib/python3.6/site-packages/qtconsole/frontend_widget.py", line 70, in highlightBlock
    if not self.highlighting_on:
AttributeError: 'FrontendHighlighter' object has no attribute 'highlighting_on'
```
I am guessing that this happens because some object which was not created by `FrontendHighlighter.__init__()` ends up calling FrontendHighlighter.highlightBlock(), but the `highlighting_on` attribute does not actually exist on that object. This PR should fix the problem.


ccordoba12:
Thanks!

PR #303, opened 2018-10-08
Update installation instructions for pyqt

2DSharp:
Instead of having debian package installation instructions (which are already available in the qtconsole docs), the PyQT docs have been linked for alternative ways to install pyqt. Links to both pyqt4 and pyqt5 source packages for Windows/MacOS/Linux have been added.


willingc:
Congrats for the contribution @2DSharp. Great work at the PyCon India 2018 Sprints.

PR #311, opened 2018-10-29
PR: Improve auto-scrolling during execution

jnsebgosselin:
Fixes https://github.com/spyder-ide/spyder/issues/3504

- Auto-scroll the console only if we are already at the bottom of the feed.

- Auto-scrolling is activated again if the user goes back at the end of the feed during execution.

- Make the auto-scrolling smoother (remove the jittering of the last line and of the scrollbar position)

<table>
  <tr>
    <td align="center" bgcolor=white></td>
    <td align="center"></td>
  </tr>
  <tr>
    <td align="center"><b>Before </b></td>
    <td align="center"><b>After </b></td>
  </tr>
</table>


ccordoba12:
Thanks @jnsebgosselin! I really like this and the changes are very simple.

@wmvanvliet, what do you say?

ccordoba12:
@jnsebgosselin, what if we don't auto-scrolling by default and just activate it when the output is larger than `buffer_size`?

jnsebgosselin:
> 
> 
> @jnsebgosselin, what if we don't do auto-scrolling by default and just activate it when the output is larger than `buffer_size`?

Sorry Carlos, I do not understand what you mean by this.

jnsebgosselin:
> 
> 
> > @jnsebgosselin, what if we don't do auto-scrolling by default and just activate it when the output is larger than `buffer_size`?
> 
> Sorry Carlos, I do not understand what you mean by this.

I  mean, I understand what you mean, I don't understand why you would want to do that.

ccordoba12:
> I  mean, I understand what you mean, I don't understand why you would want to do that.

Well, just in case you'd like to see the output shown at the beginning, instead of scrolling up by hand. But that's for another PR, I think.

PR #331, opened 2019-03-18
Fix cursor not moving while debugging

### LINKED ISSUE #330: Can not move cursor during debugging
While debugging, The cursor stays at the last possible position. I see it moving but it comes back immediately.
To reproduce:
- `%debug range(1)` 
-  `a b d <Left> c`

I would expect `abcd` but I get `abdc`

impact27:
Stop the cursor from moving at the end while debugging.

Fixes #330.

### DIFF
--- a/qtconsole/console_widget.py
+++ b/qtconsole/console_widget.py
@@ -2086,6 +2086,11 @@ class ConsoleWidget(MetaQObjectHasTraits('NewBase', (LoggingConfigurable, superQ
         else:
             cursor.insertText(text)
         cursor.endEditBlock()
+        
+        if self._executing and end_doc_pos - end_scroll_pos <= 1:
+            end_scroll = (self._control.verticalScrollBar().maximum()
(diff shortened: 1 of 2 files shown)


impact27:
I am probably breaking something by doing that, but I'm not sure what.

ccordoba12:
This was introduced by @jnsebgosselin here

https://github.com/jupyter/qtconsole/commit/d8a4810f4046e3cde098f0e1e336b447ce2b5b20

so I think you need to find a workaround.

impact27:
Now it won't touch the cursor but will just scroll down, which I think was the intended behaviour

ccordoba12:
Thanks for the update @impact27!

@jnsebgosselin, could you take a quick look at this one? Thanks!

jnsebgosselin:
Sure, I'll do it sometime this week.

ccordoba12:
Merging because the change seems pretty minimal and it fixes an important regression in 4.4.3

PR #349, opened 2019-07-02
Scroll down for output from remote command

### LINKED ISSUE #347: console does not scroll down when output is produced by a remote command
qtconsole_scrolling_issue

I've actually been looking for a possible solution as it seems an easy fix. However I'm a little lost as I'm not too familiar with the code. I can have a look if you can give me some starting point.

ajtam:
Fixes #347 

Explanation
Qtconsole treats commands from an external frontend (remote) differently than commands done locally. However all remote commands go through the function "_append_custom" where I added the line to scroll to the end of the console.

scroll_output

### DIFF
--- a/qtconsole/console_widget.py
+++ b/qtconsole/console_widget.py
@@ -944,7 +944,7 @@ class ConsoleWidget(MetaQObjectHasTraits('NewBase', (LoggingConfigurable, superQ
 
         # Perform the insertion.
         result = insert(cursor, input, *args, **kwargs)
-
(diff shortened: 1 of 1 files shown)


ccordoba12:
Thanks!

ccordoba12:
Thanks!

ccordoba12:
Thanks!

ccordoba12:
Thanks!

tarasowski:
I've tried to replicate the same example with the version 4.5.5, the auto-scroll is not working for charts, while for everything else it does. I'm using MacOS Catalina 10.15

PR #357, opened 2019-08-05
Add an option to configure scrollbar visibility

seeralans:
Added the visibility of the side scrollbar as a configurability option.


wmvanvliet:
Just gave this a try. Works great! +1 for merge from me.

PR #359, opened 2019-08-07
PR: Don't scroll down for standard use

### LINKED ISSUE #351: Prompt automatically scrolling down to the bottom of the window on execution
I would like the prompt to stay where it is while executing. Rapidly auto-scrolling to the bottom gets annoying/jarring quickly. Any way to do it?  (Ubuntu 19.10, QtConsole 4.5.1 running Python 3.7.3)

dalthviz:
Fixes #351 

Use `self._reading` as part of the validation to ensure scroll behavior only while debugging

### DIFF
--- a/qtconsole/console_widget.py
+++ b/qtconsole/console_widget.py
@@ -2086,7 +2086,10 @@ class ConsoleWidget(MetaQObjectHasTraits('NewBase', (LoggingConfigurable, superQ
             cursor.insertText(text)
         cursor.endEditBlock()
 
-        if self._executing and end_doc_pos - end_scroll_pos <= 1:
+        if (self._executing and end_doc_pos - end_scroll_pos <= 1
+                and self._reading):
+            # Only apply this behavior while debugging.
+            # See issue #351
             end_scroll = (self._control.verticalScrollBar().maximum()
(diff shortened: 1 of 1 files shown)


ccordoba12:
@impact27, is this fix ok for you?

@espdev, could you test this PR to see if it fixes the regression discussed in #351?

impact27:
The relevant PR are: 
#311
#331

 - `self._reading` is not really a reliable way to determine if we are debugging. We could be debugging but not executing, or reading but not debugging.

The requirements for fixing this problem are:
 - The auto-scrolling should only be active when the user is NOT inputing strings (Either with pdb or with the regular input)
 - Auto-scroll the console only if we are already at the bottom of the feed.
 - Auto-scrolling is activated again if the user goes back at the end of the feed during execution.

impact27:
Also:
 - Only scroll if there is an output, ot if space is needed for a new prompt.

impact27:
This should not be active only while debugging.

impact27:
I propose an alternative solution here: #360

ccordoba12:
@impact27, thanks for taking a look at this one and proposing an alternative solution!

PR #360, opened 2019-08-08
PR: Fix scrolling issues

### LINKED ISSUE #351: Prompt automatically scrolling down to the bottom of the window on execution

impact27:
Fixes #351
Before:
before

After:
after

### DIFF
--- a/qtconsole/console_widget.py
+++ b/qtconsole/console_widget.py
@@ -944,7 +944,6 @@ class ConsoleWidget(MetaQObjectHasTraits('NewBase', (LoggingConfigurable, superQ
 
         # Perform the insertion.
         result = insert(cursor, input, *args, **kwargs)
-        self._control.moveCursor(QtGui.QTextCursor.End)
         return result
 
     def _append_block(self, block_format=None, before_prompt=False):
@@ -2089,7 +2088,9 @@ class ConsoleWidget(MetaQObjectHasTraits('NewBase', (LoggingConfigurable, superQ
         if self._executing and end_doc_pos - end_scroll_pos <= 1:
             end_scroll = (self._control.verticalScrollBar().maximum()
                           - self._control.verticalScrollBar().pageStep())
-            self._control.verticalScrollBar().setValue(end_scroll)
+            # only go up
+            if end_scroll > self._control.verticalScrollBar().value():
+                self._control.verticalScrollBar().setValue(end_scroll)
 
     def _insert_plain_text_into_buffer(self, cursor, text):
         """ Inserts text into the input buffer using the specified cursor (which
--- a/qtconsole/console_widget.py
+++ b/qtconsole/console_widget.py
@@ -2088,7 +2088,7 @@ class ConsoleWidget(MetaQObjectHasTraits('NewBase', (LoggingConfigurable, superQ
         if self._executing and end_doc_pos - end_scroll_pos <= 1:
             end_scroll = (self._control.verticalScrollBar().maximum()
                           - self._control.verticalScrollBar().pageStep())
-            # only go up
+            # Only scroll down
             if end_scroll > self._control.verticalScrollBar().value():
                 self._control.verticalScrollBar().setValue(end_scroll)
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,7 +5,7 @@ sudo: false
 env:
   global:
     - CONDA_DEPENDENCIES='pyqt traitlets ipython_genutils jupyter_core jupyter_client pygments ipykernel pytest pytest-cov mock'
-    - PIP_DEPENDENCIES='coveralls pytest-xvfb'
+    - PIP_DEPENDENCIES='coveralls pytest-xvfb pytest-qt pytest-ordering'
 
 matrix:
   include:
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,7 +5,7 @@ sudo: false
 env:
   global:
     - CONDA_DEPENDENCIES='pyqt traitlets ipython_genutils jupyter_core jupyter_client pygments ipykernel pytest pytest-cov mock'
-    - PIP_DEPENDENCIES='coveralls pytest-xvfb pytest-qt pytest-ordering'
+    - PIP_DEPENDENCIES='coveralls pytest-xvfb pytest-qt'
 
 matrix:
   include:
--- a/qtconsole/tests/test_00_console_widget.py
+++ b/qtconsole/tests/test_00_console_widget.py
@@ -31,7 +31,6 @@ def qtconsole(qtbot):
     return console
 
 
-@pytest.mark.first
 @pytest.mark.parametrize(
     "debug", [True, False])
 def test_scroll(qtconsole, qtbot, debug):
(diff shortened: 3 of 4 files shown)


impact27:
I still need some tests. I want to check the position of `w._control.verticalScrollBar().value()`.
- Running a single command with enough prompts that it could scroll down to check it stays in place

```python
import time
for i in range(10000):
    print(i)
    time.sleep(.1)
```
- run this loop and check:
     - The verticalScrollBar scrolls with the output
     - If I move up, the verticalScrollBar stops following the output
     - if I move back down, the verticalScrollBar follows the output again

I am unsure how to do that, are there tests where a console is actually opened?

qtconsole/console_widget.py
```diff
@@ -2089,7 +2088,9 @@ def _insert_plain_text(self, cursor, text, flush=False):
         if self._executing and end_doc_pos - end_scroll_pos <= 1:
             end_scroll = (self._control.verticalScrollBar().maximum()
                           - self._control.verticalScrollBar().pageStep())
-            self._control.verticalScrollBar().setValue(end_scroll)
+            # only go up
```
wmvanvliet: don't you mean scoll *down*?

impact27: Well the text goes up

impact27: But I can change the comment

ccordoba12:
> I am unsure how to do that, are there tests where a console is actually opened?

Not here, but you can find how to do that in this file:

https://github.com/spyder-ide/spyder-kernels/blob/9d62e2e891fb133b086ba42d6f8a3e0c21277b45/spyder_kernels/ipdb/tests/test_matplotlib.py

impact27:
I added test but I don't know exactly how to run pytest while the rest is in unittest. Could somebody fix the tests? They all run individually on my machine.

qtconsole/tests/test_console_widget.py
```diff
@@ -415,35 +417,140 @@ def is_complete(self, source):
...
-        # Initialize widget
-        responses = []
-        w = ConsoleWidget()
-        w._append_plain_text('Header\n')
-        w._prompt = 'prompt>'
-        w._show_prompt()
-        w.kernel_client = TestIPyKernelClient()
-        control = w._control
+@pytest.fixture
+def qtconsole():
```
ccordoba12: Please move this fixture to the beginning of the file, make it depend on `qtbot` and use `qtbot` as in the file I pointed you to.

qtconsole/tests/test_console_widget.py
```diff
@@ -415,35 +417,140 @@ def is_complete(self, source):
...
+    """Qtconsole fixture."""
+    # Create a console with the ipdb kernel
```
ccordoba12: This comments doesn't apply here.

qtconsole/tests/test_console_widget.py
```diff
@@ -415,35 +417,140 @@ def is_complete(self, source):
...
+def test_scroll(qtconsole):
+    """
+    Make sure the scrolling works.
+    """
+    window = qtconsole.window
+    shell = window.active_frontend
+    wait_loop = QtCore.QEventLoop()
+    shell.executed.connect(wait_loop.quit)
+    control = shell._control
+    scroll_bar = control.verticalScrollBar()
```
ccordoba12: The previous six lines are completely unnecessary by using `qtbot`. Please remove them and replace them with the code I used in the referenced file.

qtconsole/tests/test_console_widget.py
```diff
@@ -415,35 +417,140 @@ def is_complete(self, source):
...
+
+    def QtWait(time):
+        wait_timeout = QtCore.QTimer()
+        wait_timeout.setSingleShot(True)
+        wait_timeout.timeout.connect(wait_loop.quit)
+        wait_timeout.start(time * 1000)
+        wait_loop.exec_()
+        wait_timeout.timeout.disconnect(wait_loop.quit)
```
ccordoba12: `qtbot` was a `wait` method. Please use it instead of reinventing the wheel here.

qtconsole/tests/test_console_widget.py
```diff
@@ -415,35 +417,140 @@ def is_complete(self, source):
...
+
+    QtWait(1)
+
+    assert scroll_bar.value() == 0
+    # Create a bunch of inputs
+    for i in range(20):
+        QTest.keyClicks(control, 'a = 1')
+        QTest.keyClick(control, QtCore.Qt.Key_Enter,
+                       modifier=QtCore.Qt.ShiftModifier)
```
ccordoba12: `qtbot` also has a method to do key clicks. Please use it instead of this.

qtconsole/tests/test_console_widget.py
```diff
@@ -415,35 +417,140 @@ def is_complete(self, source):
...
+
+    # reset position
+    prev_position = scroll_bar.maximum() - (scroll_bar.pageStep() * 8) // 10
+    scroll_bar.setValue(prev_position)
+
+    QtWait(1)
+    assert scroll_bar.value() > prev_position
+
+
+def test_debug(qtconsole):
```
ccordoba12: The same comments I left for `test_scroll` apply to this one.

ccordoba12:
> I added test but I don't know exactly how to run pytest while the rest is in unittest

I don't understand what you mean by this. Pytest runs all our tests, as can be seen here:

https://github.com/jupyter/qtconsole/blob/master/.travis.yml#L21

impact27:
I had to run the new tests first for some reason, but at least it works. I would be grateful if someone can explain why I needed to do that.

qtconsole/tests/test_aaconsole_widget.py
```diff
@@ -10,12 +10,139 @@
...
+    qtbot.waitUntil(lambda: shell._prompt_html is not None,
+                    timeout=SHELL_TIMEOUT)
+
+    assert scroll_bar.value() == 0
+    # Create a bunch of inputs
+    for i in range(20):
+        with qtbot.waitSignal(shell.executed):
+            QTest.keyClicks(control, 'a = 1')
+            QTest.keyClick(control, QtCore.Qt.Key_Enter,
+                           modifier=QtCore.Qt.ShiftModifier)
```
ccordoba12: You're still using `QTest.keyClicks` instead of `qtbot.keyClicks`. Please use the later.

ccordoba12: But only in your new tests, I mean.

qtconsole/console_widget.py
wmvanvliet: I find `# Only scroll down` easier to understand. Sorry for the nitpick ;)

ccordoba12:
I think this is almost ready. One last comment: is it still necessary to rename `test_console_widget`? I thought that by using pytest-ordering that wouldn't be necessary.

impact27:
I didn't manage to make it work, I think it is only reordering tests in a same file.

ccordoba12:
Then there's no need to use pytest-ordering, right?

I also think it's better to rename `test_console_widget.py` to `00_test_console_widget.py`, to avoid future files to cause failures.

impact27:
`test_00_console_widget.py` so pytest still runs it.

ccordoba12:
Yep, right.

PR #366, opened 2019-08-20
Fix emoji highlight

### LINKED ISSUE #365: String highlighting is broken with emoji.
`a = '🚫'` is highlighted as:
Screenshot 2019-08-19 at 18 40 08
I think this is caused by the cursor position in Qt not corresponding to the character count. For example, `🚫` is a single character. In python, `len('🚫') = 1`. But Qt doesn't count characters, but QChar, and the UTF16 encoding needs 2 QChar to represent `🚫`.

From https://doc.qt.io/qt-5/qtextcursor.html:

Note: The "characters" in this case refer to the string of QChar objects, i.e. 16-bit Unicode characters, and the position is considered an index into this string. This does not necessarily correspond to individual graphemes in the writing system, as a single grapheme may be represented by multiple Unicode characters, such as in the case of surrogate pairs, linguistic ligatures or diacritics.

impact27:
fixes #365

### DIFF
--- a/qtconsole/qstringhelpers.py
+++ b/qtconsole/qstringhelpers.py
@@ -1,0 +1,25 @@
+# -*- coding: utf-8 -*-
+#
+# Copyright © Spyder Project Contributors
+# Licensed under the terms of the MIT License
+# (see spyder/__init__.py for details)
+
+"""QString compatibility."""
+
+from spyder.py3compat import PY2
+
+
+def qstring_length(text):
+    """
+    Tries to compute what the length of an utf16-encoded QString would be.
+    """
+    if PY2:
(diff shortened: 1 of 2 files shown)


qtconsole/qstringhelpers.py
```diff
@@ -0,0 +1,25 @@
+# -*- coding: utf-8 -*-
+#
+# Copyright © Spyder Project Contributors
+# Licensed under the terms of the MIT License
+# (see spyder/__init__.py for details)
+
+"""QString compatibility."""
+
+from spyder.py3compat import PY2
```
ccordoba12: Please change this to

    PY2 = sys.version[0] == '2'

PR #368, opened 2019-08-26
Scroll when adding text even when not executing.

impact27:
If text is added to the console, even if the console is not executing, the console should scroll.


ccordoba12:
Please post screenshots of before/after this PR to see if it's really important to make this change or not.

impact27:
It's not really important to make this change. I noticed this when I made an error while setting _executing in spyder. I remember some instances of messges being printed while the prompt is visible (So `_reading` and not `_executing`). For example when threads print to the console. In this case, the printing threads could hide the prompt.

ccordoba12:
Ok, I trust your judgment about this then.

PR #369, opened 2019-08-28
Allow text to be added before the prompt while autocompleting

impact27:
Autocomplete will fail it text is added to the console between the autocomplete request and the autocomplet reply, even if this text is added before the prompt. I would expect this to fail only if the prompt is changed. Therefore, instead of checking the absolute position, this checks the prompt buffer and the position in the prompt buffer.

### DIFF
--- a/qtconsole/frontend_widget.py
+++ b/qtconsole/frontend_widget.py
@@ -157,7 +157,8 @@ class FrontendWidget(HistoryConsoleWidget, BaseFrontendMixin):
     exit_requested = QtCore.Signal(object)
 
     _CallTipRequest = namedtuple('_CallTipRequest', ['id', 'pos'])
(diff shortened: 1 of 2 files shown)


ccordoba12:
> Autocomplete will fail it text is added to the console between the autocomplete request and the autocomplet reply, even if this text is added before the promp

When does this case appear?

impact27:
This appears sometimes while debugging. Sometimes the kernel will send the prompt before sending the cout content. This is why the option of adding text before the prompt exist. If the user manages to ask for completion before the cout text is displayed, and the reply comes after, then this is triggered.

impact27:
Of course, this is almost impossible for a user to do that, but CI can trigger this.

ccordoba12:
Ok, I see. Thanks for the clarification.

PR #371, opened 2019-08-28
Set console to read only after input

### LINKED ISSUE #370: Disable undo and edit after input.
1) Run this code:
Screenshot 2019-08-28 at 16 17 21
 2) Enter any string:
Screenshot 2019-08-28 at 16 18 14
3) Ctrl + Z
Screenshot 2019-08-28 at 16 18 22
4) backspace multiple times
Screenshot 2019-08-28 at 16 18 28

impact27:
Fixes #370.

Could that be a new version of qtconsole, so I can use that in https://github.com/spyder-ide/spyder/pull/10087?

### DIFF
--- a/qtconsole/tests/test_00_console_widget.py
+++ b/qtconsole/tests/test_00_console_widget.py
@@ -137,6 +137,38 @@ def test_scroll(qtconsole, qtbot, debug):
     assert scroll_bar.value() > prev_position
 
 
+@pytest.mark.skipif(sys.version[0] == '2', reason="input and py2")
+def test_input(qtconsole, qtbot):
+    """
+    Test input function
+    """
+    window = qtconsole.window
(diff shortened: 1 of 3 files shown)


qtconsole/tests/test_00_console_widget.py
```diff
@@ -137,6 +137,38 @@ def run_line(line, block=True):
     assert scroll_bar.value() > prev_position
 
 
+@pytest.mark.skipif(sys.version[0] == '2', reason="input and py2")
```
ccordoba12: This reason is not clear enough for me. Please word it better.

ccordoba12:
> Could that be a new version of qtconsole, so I can use that in spyder-ide/spyder#10087?

Sure, I'll try to release it this week, but I don't know when it's going to be available in Anaconda.

PR #373, opened 2019-09-05
fix_return

### LINKED ISSUE #372: Pressing return sometimes goes to the next line instead of running the command
Example of this issue crashing a CI: https://travis-ci.org/spyder-ide/spyder/jobs/580645019

when `_trigger_is_complete_callback` is called from `eventFilter` instead of being called from `_handle_is_complete_reply`, `complete` is set to `False` even if the command is complete. 
Screenshot 2019-09-04 at 13 05 15

impact27:
Wait 250ms for a reply from is_complete instead of giving up immediately.
fix #372


impact27:
Closing for now as it causes https://github.com/spyder-ide/spyder/issues/10161

impact27:
A different approach that doesn't trigger spyder-ide/spyder#10161

PR #375, opened 2019-09-09
Wait for a reply from is_complete instead of giving up immediately

### LINKED ISSUE #372: Pressing return sometimes goes to the next line instead of running the command

impact27:
Wait 250ms for a reply from is_complete instead of giving up immediately.
fix #372

### DIFF
--- a/qtconsole/console_widget.py
+++ b/qtconsole/console_widget.py
@@ -223,9 +223,8 @@ class ConsoleWidget(MetaQObjectHasTraits('NewBase', (LoggingConfigurable, superQ
             self.setParent(parent)
 
         self._is_complete_msg_id = None
-        self._is_complete_timeout = 0.25
+        self._is_complete_timeout = 0.1
         self._is_complete_max_time = None
-        self._is_complete_number_returns = 0
 
         # While scrolling the pager on Mac OS X, it tears badly.  The
         # NativeGesture is platform and perhaps build-specific hence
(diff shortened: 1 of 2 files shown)


impact27:
retriggering tests to make sure it is a real problem

qtconsole/console_widget.py
```diff
@@ -223,6 +223,9 @@ def __init__(self, parent=None, **kw):
             self.setParent(parent)
 
         self._is_complete_msg_id = None
+        self._is_complete_timeout = 0.25
```
ccordoba12: Could this time be reduced a bit more? Say 0.1?

qtconsole/console_widget.py
```diff
@@ -1231,10 +1242,11 @@ def callback(complete, indent):
                             try:
                                 cursor.beginEditBlock()
                                 cursor.setPosition(position)
-                                cursor.insertText('\n')
-                                self._insert_continuation_prompt(cursor)
-                                if indent:
-                                    cursor.insertText(indent)
+                                for i in range(self._is_complete_number_returns):
```
ccordoba12: What's this taking care of? I don't understand why it's needed as part of this PR.

qtconsole/console_widget.py
impact27: 0.1 works as well.

qtconsole/console_widget.py
impact27: The idea was to avoid dropping returns, but it is not necessary.

Issue #279, opened 2018-03-29
qt-console not working with pyside2


looooo:
pyside2 is already quite stable and easily installable with conda / conda-forge. Maybe someone can have a look at this.
```
 File "<frozen importlib._bootstrap>", line 219, in _call_with_frames_removed
  File "~/projects/freecad-ipython/freecad/ipython_integration/init_gui.py", line 34, in <module>
    console = put_ipy(dock_widget, mw)
  File "~/projects/freecad-ipython/freecad/ipython_integration/init_gui.py", line 22, in put_ipy
    widget = RichJupyterWidget(parent=parent)
  File "~/conda/envs/freecad/lib/python3.6/site-packages/qtconsole/rich_jupyter_widget.py", line 53, in __init__
    super(RichJupyterWidget, self).__init__(*args, **kw)
  File "~/conda/envs/freecad/lib/python3.6/site-packages/qtconsole/jupyter_widget.py", line 115, in __init__
    super(JupyterWidget, self).__init__(*args, **kw)
  File "~/conda/envs/freecad/lib/python3.6/site-packages/qtconsole/frontend_widget.py", line 181, in __init__
    super(FrontendWidget, self).__init__(*args, **kw)
  File "~/conda/envs/freecad/lib/python3.6/site-packages/qtconsole/history_console_widget.py", line 28, in __init__
    super(HistoryConsoleWidget, self).__init__(*args, **kw)
  File "~/conda/envs/freecad/lib/python3.6/site-packages/qtconsole/console_widget.py", line 289, in __init__
    self.reset_font()
  File "~/conda/envs/freecad/lib/python3.6/site-packages/qtconsole/console_widget.py", line 784, in reset_font
    font.setPointSize(QtGui.QApplication.instance().<class 'AttributeError'>: 'PySide2.QtCore.QCoreApplication' object has no attribute 'font'

```

ps.: I am using this dependencies:
```
pyqt                      5.6.0                    py36_4    conda-forge
qt                        5.6.2                         7    conda-forge
qtconsole                 4.3.1                    py36_0    conda-forge
pyside2                   5.6.0a1                  py36_0    conda-forge
ipython                   6.2.1                    py36_1    conda-forge
jupyter                   1.0.0                      py_1    conda-forge
jupyter_client            5.2.3                    py36_0    conda-forge
jupyter_console           5.2.0                    py36_0    conda-forge
jupyter_core              4.4.0                      py_0    conda-for
```

[...]

PR #381, opened 2019-09-22
Use qtpy instead of rolling custom loader

hmaarrfk:
Use qtpy for Qt backend selector

I think I also happened to kill support for PySide (1). Is it time to drop support for 2.7???

closes: https://github.com/jupyter/qtconsole/issues/279


qtconsole/qt.py
```diff
@@ -1,23 +1,5 @@
 """ A Qt API selector that can be used to switch between PyQt4/5 and PySide.
```
ccordoba12: I think this submodule should be removed and `qtpy` used directly wherever is needed.

ccordoba12:
> I think I also happened to kill support for PySide (1)

`qtpy` still supports it.

> Is it time to drop support for 2.7???

Not yet, but in any case it shouldn't be part of this PR.

hmaarrfk:
i don't really have a plan for getting past the 2.7 test fail.

Do you know it is happening?

qtconsole/frontend_widget.py
```diff
@@ -169,19 +168,10 @@ def _lexer_default(self):
 
     def __init__(self, local_kernel=_local_kernel, *args, **kw):
         super(FrontendWidget, self).__init__(*args, **kw)
-        # FIXME: remove this when PySide min version is updated past 1.0.7
-        # forcefully disable calltips if PySide is < 1.0.7, because they crash
```
hmaarrfk: @ccordoba12 this is why I think I killed PySide support.

hmaarrfk: I guess it is reasonable to expect people not to be using a version from 2011??

https://pypi.org/project/PySide/

hmaarrfk:
Is the `test_debug` known to be inconsistent? 

I'm getting inconsistent results locally

https://travis-ci.org/jupyter/qtconsole/jobs/588225647#L611

One set failed on python 3.6, while the other failed on python 2.7.

ITAYC0HEN:
Looks good. Rebase is required

hmaarrfk:
Lets see what happens.

ITAYC0HEN:
Looks green :)

CONTRIBUTING.md
```diff
@@ -11,7 +11,7 @@ Use Conda to install dependencies and activate the development environment.
 ```
 conda create -n qtdev python=3
 conda activate qtdev
-conda install pyqt
```
ccordoba12: `qtpy` doesn't bring `pyqt`, so this needs to be

    conda install pyqt qtpy

examples/inprocess_qtconsole.py
```diff
@@ -10,7 +10,7 @@
 """
 
 
-from qtconsole.qt import QtGui
+from qtpy import QtGui
```
ccordoba12: This needs to be

    from qtpy import QtWidgets

to use the right import below.

qtconsole/frontend_widget.py
ccordoba12: Agreed.

hmaarrfk:
Thanks, hopefully all passes

ccordoba12:
> Is it time to drop support for 2.7???

It is. But let's release this in the last Python 2 compatible version. Afterwards you could send us a Python 3 migration PR

ITAYC0HEN:
Congratulations! Good work, folks :)

ccordoba12:
I'll try to release this during the next week.

hmaarrfk:
this, using qtpy as a unified backend for my software is definitely really nice!

ITAYC0HEN:
> I'll try to release this during the next week.

Hey @ccordoba12! :) Can you please release upstream to pypi so we can use it from pip? :)

ccordoba12:
Done.

PR #394, opened 2019-11-16
Prevent completion strip if it refers to a file

dalthviz:
A preview:

qc

Fixes spyder-ide/spyder#10692

Pinging @ccordoba12 just in case


ccordoba12:
Thanks a lot @dalthviz for this one! There are two things here:

1. Tests are failing now.
2. Could we increase the widget width to see a bit more content or add an horizontal scrollbar to it? In you screenshot above the file extension of `Sin título<n>.py` is not visible at some point.

dalthviz:
A new preview:

qc

PR #417, opened 2020-04-08
Replace misuses of QtGui

dalthviz:
Fixes spyder-ide/spyder#12210


ccordoba12:
@dalthviz, please make an audit of all the places where we're using `QtGui` here, to see if they need to be replaced by `QtWidgets` or not.

That way we'll avoid making this kind of minor fixes several times in the future.

ccordoba12:
Is this one ready @dalthviz?

dalthviz:
yep @ccordoba12 :+1:

PR #419, opened 2020-04-28
Fix a TypeError in CallTipWidget

### LINKED ISSUE #383: Crash when clicking a calltip away
## Context

Hello all, I've been contributing to the [Mu Editor](https://github.com/mu-editor/mu) -- a beginner friendly Python editor -- which builds its GUI on top of Qt and, among other things, uses `qtconsole` for its Python REPL pane. Before anything else, thanks for sharing `qtconsole`. :)

## The failure

Out of luck I detected [a critical bug](https://github.com/mu-editor/mu/issues/934) that leads to a Mu crash that seems to point to a misbehaviour in `qtconsole` (or, maybe, somewhere else).

In short, when a call tip is visible, clicking with the mouse somewhere within the `qtconsole` widget in an attempt to close the call tip, leads to an exception like:
```
  File ".../qtconsole/call_tip_widget.py", line 89, in hideEvent
    self._cursor_position_changed)
TypeError: 'method' object is not connected
```

The interesting (sad?) part is that such exception leads to a hard exit with PyQt5 but, apparently, not so with PySide2.

## Steps to reproduce

From my testing, with both the Mu Editor and a "plain" `jupyter-qtconsole`, it only seems to fail this way on macOS. Somehow Windows and Linux don't behave exactly the same.

* macOS 10.12.6
* CPython 3.6.8 package from www.python.org installed.

```
$ git clone https://github.com/jupyter/qtconsole/
$ cd qtconsole
$ python3.6 -m venv venv
$ . venv/bin/activate
(venv) $ pip install -e .
(venv) $ pip install pyqt5
(venv) $ jupyter-qtconsole
```

Then, in the GUI `qtconsole`:
```
In [1]: print(
```

At this point, a call tip for `print` pops up. Now, mouse click anywhere on the `qtconsole` widget.

The traceback is printed out and the process terminates with:
```
  File ".../qtconsole/call_tip_widget.py", line 89, in hideEvent
    self._cursor_position_changed)
TypeError: 'method' object is not connected
Abort trap: 6
(venv) $
```

## What I tried

Investigated a bit and found out:

* Running the above with PySide2 instead of PyQt5 produces the same exception and traceback, but doesn't crash the running process. Instead, PySide2 seems to handle it better, printing out `RuntimeError: Failed to disconnect signal cursorPositionChanged().`, after the traceback.
* The failure seems to result from the fact that `hideEvent` in `CallTipWidget` is being called twice -- add a `print` to `showEvent` and `hideEvent` to see that.
* The following platform differences are interesting, not sure if relevant. Whenever the call tip is visible:
  * On Windows and Linux, clicking somewhere "below" the blinking cursor has no effect. Only clicking somewhere "above" does make the call tip go away.
  * On macOS, clicking anywhere leads to the observed failure.

## Eventually related

I found [this Spyder issue](https://github.com/spyder-ide/spyder/issues/10050) showing a similar exception, apparently on macOS, given the file path, which was fixed [with this PR here](https://github.com/jupyter/qtconsole/pull/367). I'm not sure that PR is the "correct" / "best" fix and I wonder if in that case `hideEvent` wasn't being called twice, too. If so, that might be the root cause for the failure.

## Wrap up

Again, thanks for `qtconsole`. I'm not familiar with its design or codebase, much less with PyQt5's. I'll be happy to collaborate in order to see this failure understood and fixed and I'm not sure this is a `qtconsole` issue, or something Qt related. Thanks in advance for your support in this. :)

rayosborn:
This is a fix for jupyter/qtconsole#383. It applies the same modification to the `CallTipWidget` `hideEvent` function as that used to fix `CompletionWidget` in #367. 

Fixes #383.

### DIFF
--- a/qtconsole/call_tip_widget.py
+++ b/qtconsole/call_tip_widget.py
@@ -85,8 +85,12 @@ class CallTipWidget(QtWidgets.QLabel):
         """ Reimplemented to disconnect signal handlers and event filter.
         """
         super(CallTipWidget, self).hideEvent(event)
-        self._text_edit.cursorPositionChanged.disconnect(
-            self._cursor_position_changed)
+        # This fixes issue jupyter/qtconsole#383
+        try:
+            self._text_edit.cursorPositionChanged.disconnect(
+                self._cursor_position_changed)
+        except TypeError:
+            pass
         self._text_edit.removeEventFilter(self)
 
     def leaveEvent(self, event):


wmvanvliet:
Yeah, this would work. However, it would be even better if we could prevent the error from occuring in the first place, rather than just swallowing it and pretend it didn't happen.

rayosborn:
I have no problem with using another solution as long as the fix gets into the next release.

ccordoba12:
@rayosborn, please merge with master to get the fix to our tests.

rayosborn:
I'm not sure I understand the request. Should I merge with my fork's master and then reissue the PR?

ccordoba12:
No, please run these commands instead:

```
git remote add upstream https://github.com/jupyter/qtconsole.git
git checkout master
git fetch upstream
git merge upstream/master
git checkout fix-calltipwidget
git merge master
git push origin fix-calltipwidget
```

PR #420, opened 2020-04-28
PR: Verify if event has attribute button

steff456:
Verify if the event has button as an attribute before calling `event.button()`

Fixes https://github.com/spyder-ide/spyder/issues/12524


ccordoba12:
@steff456, please merge with master to get the fix to our tests.

PR #421, opened 2020-05-04
PR: Fix completion widget text for paths and files

dalthviz:
Fixes spyder-ide/spyder#12444

### DIFF
--- a/qtconsole/completion_widget.py
+++ b/qtconsole/completion_widget.py
@@ -116,7 +116,7 @@ class CompletionWidget(QtWidgets.QListWidget):
             else:
                 list_item = QtWidgets.QListWidgetItem()
                 list_item.setData(QtCore.Qt.UserRole, item)
-                list_item.setText(item.split('.')[-1])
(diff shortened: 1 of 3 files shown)


qtconsole/completion_widget.py
```diff
@@ -116,7 +116,8 @@ def show_items(self, cursor, items, prefix_length=0):
             else:
                 list_item = QtWidgets.QListWidgetItem()
                 list_item.setData(QtCore.Qt.UserRole, item)
-                list_item.setText(item.split('.')[-1])
+                # Need to split to only show last element of a dot completion
+                list_item.setText(item.split(".")[-1])
```
wmvanvliet, line 120: why change the quote type?

dalthviz: Ups my bad, by mistake I changed this line in the first commit and when I re-added it changed the quote type. Would you like better the `"` instead of the `'`? (in terms of functionality I think they work the same in this case).

PR #422, opened 2020-05-05
PR: Fix Python 2 tests

dalthviz:
@ccordoba12 ~~at the end I removed the testing in Python 2.7 (no idea what is causing the memory segfault, the only difference I saw in the logs was the change from Python `2.7.17` to Python `2.7.18` and that the Qt version changed from `5.9.7` to `5.9.4`)~~ seems like using Python 2.7.17 prevents the tests segfault in Python 2

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,7 +11,7 @@ matrix:
   include:
     - env: PYTHON_VERSION=3.7
     - python: "3.6"
-    - python: "2.7"
+    - python: "2.7.17"
 
 install:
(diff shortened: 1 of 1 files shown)


ccordoba12:
> I saw in the logs was the change from Python 2.7.17

Could you our tests to this version then to see if they pass with it? If not, then I agree we should remove them.

dalthviz:
@ccordoba12 I think this is ready :+1:

PR #427, opened 2020-06-28
Flag to control whether code from other clients is shown

randomgeek78:
When working with several remote clients connected to the same kernel, we may want to only display the execution result on a client but not the execution input code as that is available to use elsewhere.

This update keeps the workspace clean and focussed. It mimics the behavior of the ipython magics `%paste -q` where only the result is shown but the pasted code is not displayed.

This update is especially nice when working with an editor like `neovim` driving the development through plugins like `nvim-ipy`.


wmvanvliet:
As developer of [jupyter-vim](https://github.com/jupyter-vim/jupyter-vim) I feel your plight.

Ideally, I would like the editor to be able to control the verbosity for each execution message individually, so users can configure things just as they like it. A good default would for example be to copy MATLAB's approach: executing single lines will cause them to be echo-ed, but executing a code cell will not.

However, to make this possible, this should really be solved at the kernel level instead of at the client. The good news is that the functionality is already sort of there. The kernel protocol has a [`silent` flag](https://jupyter-client.readthedocs.io/en/latest/messaging.html#execute) that clients should respect. Unfortunately, setting this flag on an execute message will suppress all echo-ing, also that of the result. We could try and get a new value added for this flag so that only the input-echo is disabled.

For the time being, we could also experiment with having nvim first send everything silently, followed by a non-silent `print(_)` afterwards to echo the result.

WDYT?

1 + 1

randomgeek78:
@wmvanvliet Thanks for your thoughtful comments. I agree that to some, this may seem like near OCR but these small things matter a lot when you are working with it in a tight loop. For now, these changes that I made are working for me. I have not contributed much earlier so don't know how to push for this functionality. I look forward to more progress on this and a chance to help as needed.

PR #430, opened 2020-07-16
Replace qApp with QApplication.instance()

dimitriepirghie:
In newer versions of __PySide2__ the attribute _qApp_ of module _QtWidgets_ is removed.
This PR changes all the occurrences of _qApp_ with _QApplication.instance()_
As this attribute doesn't exist on newer versions of PySide2 it causes a crash on my application.
This doesn't affect previous versions of PySide2 as _qApp_ attribute is the same as _QApplication.instance()_, verified on PySide2 5.11.2.

Crash

Tested on __PySide2 5.15.0 LTS__.


ccordoba12:
Looks good to me, thanks @dimitriepirghie!

PR #436, opened 2020-08-17
Remove u-string prefixes

rahulporuri:
The u-string prefixes are redundant and unnecessary on python 3. Note that this was split into a standalone PR given the volume of changes.


ccordoba12 (approved):
Thanks @rahulporuri!

PR #438, opened 2020-08-28
Change font width calculation to use horizontalAdvance

### LINKED ISSUE #437: Incorrect character width calculation
Hi,

In [console_widget.py#L491](https://github.com/jupyter/qtconsole/blob/master/qtconsole/console_widget.py#L491) the character width is calculated using `boundingRect`. In my system (Windows 10, python 3.8.5, pyqt5  5.15.0) the result of `boundingRect` `width()` is not consistent (even though the font is monospaced).

```python
print(' ', font_metrics.boundingRect(' ').width(), font_metrics.horizontalAdvance(' '), file=f)
for c in range(ord('a'), ord('z') + 1):
  print(chr(c), font_metrics.boundingRect(chr(c)).width(), font_metrics.horizontalAdvance(chr(c)), file=f)
```

```
  1 6 <- ' ' character
a 5 6
b 6 6
c 5 6
d 5 6
e 5 6
f 5 6
g 5 6
h 5 6
i 5 6
j 4 6
k 6 6
l 5 6
m 6 6
n 5 6
o 6 6
p 6 6
q 5 6
r 5 6
s 5 6
t 5 6
u 5 6
v 6 6
w 6 6
x 5 6
y 6 6
z 5 6
```

This results in the wrong window width. What works is to use `horizontalAdvance` which is the alternative to the deprecated `width` method (https://doc.qt.io/qt-5/qfontmetricsf-obsolete.html#width-1).

char101:
Fixes #437.

### DIFF
--- a/qtconsole/console_widget.py
+++ b/qtconsole/console_widget.py
@@ -488,7 +488,7 @@ class ConsoleWidget(MetaQObjectHasTraits('NewBase', (LoggingConfigurable, superQ
         # a fudge factor of one character here.
         # Note 2: QFontMetrics.maxWidth is not used here or anywhere else due
         # to a Qt bug on certain Mac OS systems where it returns 0.
-        width = font_metrics.horizontalAdvance(' ') * self.console_width + margin
+        width = self._get_font_width() * self.console_width + margin
         width += style.pixelMetric(QtWidgets.QStyle.PM_ScrollBarExtent)
 
         if self.paging == 'hsplit':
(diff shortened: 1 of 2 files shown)


qtconsole/console_widget.py
```diff
@@ -752,12 +752,17 @@ def _get_font(self):
         """
         return self._control.document().defaultFont()
 
+    def _get_font_width(self, font=None):
+        if font is None:
+            font = self.font
+        font_metrics = QtGui.QFontMetrics(font)
+        return font_metrics.horizontalAdvance(' ') if hasattr(font_metrics, 'horizontalAdvance') else font_metrics.width(' ')
```
ccordoba12, line 759: This is too much for a single line. Please use a proper `if` to organize this code better, like this:

```python
if hasattr(font_metrics, 'horizontalAdvance'):
    return font_metrics.horizontalAdvance(' ')
else:
    return font_metrics.width(' ')
```

char101: Fixed.

PR #450, opened 2020-10-25
Emit kernel_restarted after restarting kernel

impact27:
`kernel_restarted` is emitted just before restarting the kernel, which doesn't really make sense. This emits it just after restarting the kernel.

### DIFF
--- a/qtconsole/manager.py
+++ b/qtconsole/manager.py
@@ -33,6 +33,10 @@ class QtKernelManager(KernelManager, QtKernelManagerMixin):
     client_class = DottedObjectName('qtconsole.client.QtKernelClient')
     autorestart = Bool(True, config=True)
 
+    def __init__(self, *args, **kwargs):
+        super(QtKernelManager, self).__init__(*args, **kwargs)
+        self._is_restarting = False
+
     def start_restarter(self):
         if self.autorestart and self.has_kernel:
             if self._restarter is None:
@@ -41,7 +45,7 @@ class QtKernelManager(KernelManager, QtKernelManagerMixin):
                     parent=self,
                     log=self.log,
                 )
-                self._restarter.add_callback(self._handle_kernel_restarted)
+            self._is_restarting = True
             self._restarter.start()
 
     def stop_restarter(self):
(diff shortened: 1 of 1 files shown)


qtconsole/manager.py
```diff
@@ -41,13 +45,16 @@ def start_restarter(self):
...
             self._restarter.start()
 
     def stop_restarter(self):
         if self.autorestart:
             if self._restarter is not None:
                 self._restarter.stop()
 
-    def _handle_kernel_restarted(self):
-        self.kernel_restarted.emit()
+    def post_start_kernel(self, **kw):
```
ccordoba12, line 56: When is this new method called?

ccordoba12 (commented):
Thanks for this @impact27! I left a minor comment for you.

Also, do you think this could to solve the issues we have with restarting kernels in some situations in Spyder?

qtconsole/manager.py
impact27: After the kernel is started:
 https://github.com/jupyter/jupyter_client/blob/842fa46e9507d1d85aacd224d672cb624188c15f/jupyter_client/manager.py#L312

impact27:
> Thanks for this @impact27! I left a minor comment for you.
> 
> Also, do you think this could to solve the issues we have with restarting kernels in some situations in Spyder?

I hope so, I need that for https://github.com/spyder-ide/spyder/pull/14025 to work properly after a restart

qtconsole/manager.py
impact27: For ref the callback is called here: https://github.com/jupyter/jupyter_client/blob/44980c13680f4e4226cf25f199ce4e4bb6e11296/jupyter_client/restarter.py#L112
BEFORE the kernel is started

qtconsole/manager.py
```diff
@@ -33,6 +33,10 @@ class QtKernelManager(KernelManager, QtKernelManagerMixin):
     client_class = DottedObjectName('qtconsole.client.QtKernelClient')
     autorestart = Bool(True, config=True)
 
+    def __init__(self, *args, **kwargs):
+        super(QtKernelManager, self).__init__(*args, **kwargs)
```
ccordoba12, line 37:
```suggestion
        super().__init__(*args, **kwargs)
```

Our next version is going to be compatible only with Python 3.

qtconsole/manager.py
```diff
@@ -41,13 +45,16 @@ def start_restarter(self):
...
 
     def stop_restarter(self):
         if self.autorestart:
             if self._restarter is not None:
                 self._restarter.stop()
 
-    def _handle_kernel_restarted(self):
-        self.kernel_restarted.emit()
+    def post_start_kernel(self, **kw):
+        super(QtKernelManager, self).post_start_kernel(**kw)
```
ccordoba12, line 57:
```suggestion
        super().post_start_kernel(**kw)
```

ccordoba12 (approved):
Thanks @impact27!

Issue #456, opened 2020-11-24
qtconsole==5.0.0 breaks python 2.7 environments


Freyja-Folkvangr:
Hi, 
Jupyter (along with other packages) requires qtconsole, they do not specify any version so by default it installs the latest version.

qtconsole version 5.0.0 breaks python 2.7 environments when installing requirements.

I know python2.7 is deprecating, so we're currently migrating at the company but it is not done yet.

**We had this error on our CI environment**

```
    Collecting qtconsole

    Downloading qtconsole-5.0.0.tar.gz (424 kB)

    ERROR: Command errored out with exit status 1:

    Complete output (6 lines):

    Traceback (most recent call last):

      File "<string>", line 1, in <module>

      File "/tmp/pip-install-p3ALmm/qtconsole/setup.py", line 18

        print(error, file=sys.stderr)

    SyntaxError: invalid syntax

    ----------------------------------------

    ERROR: Command errored out with exit status 1: python setup.py egg_info Check the logs for full command output.
```

ccordoba12:
Duplicate of #455

PR #457, opened 2020-11-24
Add python_requires for Python 3.6+ compatibility

### LINKED ISSUE #455: QtConsole 5.0.0 does not include python_requires tags, so attempts to install on Python 2
QtConsole 5.0.0 was released on the 23rd of November, and dropped Python 2 support. However, it does not include a `python_requires` directive, so Pip under Python 2 regards QtConsole 5.0.0 as an install candidate.

The standard fix for this is to:
1. Add a `python_requires` directive to either `setup.py` (as in [Jupyter-Sphinx](https://github.com/jupyter/jupyter-sphinx/blob/6eb309d19cebf2ef4fe7dbc092d1f69783195abc/setup.py#L37)) or `setup.cfg` (as in [Jupyter-Core](https://github.com/jupyter/jupyter_core/blob/f2ae1d5aacf5a5e5b299063fb8a1359de518ed92/setup.cfg#L24))
2. Publish a new release (presumably 5.0.1), with this change
3. Pull the bad 5.0.0 release from PyPI

I'm happy to raise a pull request for step 1, if a maintainer is willing to handle steps 2 and 3.

jamespic:
This change adds a python_requires directive, which prevents installing on unsupported Python versions. This is intended to fix #455 (And its duplicate #456), once a new release has been cut, and the bad release pulled.

Fixes #455.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -66,6 +66,7 @@ setup_args = dict(
     license                       = 'BSD',
     platforms                     = "Linux, Mac OS X, Windows",
     keywords                      = ['Interactive', 'Interpreter', 'Shell'],
+    python_requires               = '>= 3.6',
     install_requires = [
         'traitlets',
         'ipython_genutils',


ccordoba12 (commented):
Thanks for your help @jamespic!

PR #462, opened 2020-12-23
PR: Remove partial prompt on copy

impact27:
When copying, qtconsole removes whole prompts, but not partial prompts. This removes partial prompts on the first line.

PR #467, opened 2021-01-12
PR: Fix edge case copy last line

impact27:
Edge case forgotten in https://github.com/jupyter/qtconsole/pull/462
When the last selected line is the last line of the console, contains a prompt, and only partially selects the prompt


ccordoba12 (approved):
Thanks @impact27!

PR #474, opened 2021-03-15
Emit kernel_restarted only after crashing

impact27:
`kernel_restarted` should only be emitted when the kernel dies and restart, but is now emitted on any restart following #450. This solves this issue, and fixes https://github.com/spyder-ide/spyder/issues/14886


ccordoba12 (approved):
Thanks @impact27!

PR #477, opened 2021-04-08
Test jupyter_client

davidbrochart:

PR #478, opened 2021-04-08
Change Empty import

davidbrochart:
This change is needed for the next release of jupyter_client, and has no impact on the current code base anyway.
See #477


ccordoba12 (approved):
Thanks @davidbrochart!

PR #481, opened 2021-04-19
Monkeypatch RegexLexer only while in use by qtconsole.

### LINKED ISSUE #388: Importing qtconsole RichJupyterWidget in IPython breaks syntax highlighting
If I import RichJupyterWidget in an IPython terminal (ie the classic ipython in-process terminal), the syntax highlighting is broken in that terminal. The most visible sign is that the highlighter's determination of what is and isn't a string breaks, and it oscillates between thinking the text inside and outside of the string delimiters are strings. Depending on the length of the input string, the final highlighting will be correct or inverted.

To reproduce, launch IPython, then run:

```python
from qtconsole.rich_jupyter_widget import RichJupyterWidget
```

(You might need to launch IPython with `--gui=qt`.)

Then type:

```python
x = 'hello I am a long string'
```

and watch the highlighting flicker. The highlighting will be exactly inverted at the end.

terminal-flicker

Carreau:
Qtconsole highlight lines one line at a time so needs to remember state.
Other code may use the lexer; So patch only when relevant.

Closes  #388

### DIFF
--- a/qtconsole/pygments_highlighter.py
+++ b/qtconsole/pygments_highlighter.py
@@ -74,7 +74,19 @@ def get_tokens_unprocessed(self, text, stack=('root',)):
 
 
 # Monkeypatch!
-RegexLexer.get_tokens_unprocessed = get_tokens_unprocessed
+from contextlib import contextmanager
+
+
+@contextmanager
+def _lexpatch():
+
+    try:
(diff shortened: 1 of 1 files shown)


Carreau:
@jni @goanpeca

goanpeca (approved):

PR #485, opened 2021-05-18
Move cursor to beginning of buffer if on the same line

impact27:
Move cursor to beginning of buffer if on the same line.
Now when pressing 'b' from

`In [2]:| aaa`

leads to:

`In [2]: aaab`

After this PR:

`In [2]: baaa`


impact27:
@ccordoba12 This is ready for review

ccordoba12 (approved):
Looks good to me, thanks @impact27!

PR #488, opened 2021-06-09
Improve handling of different keyboard combinations

### LINKED ISSUE #483: Cmd-Left and Cmd-Right not as expected on MacOS
On MacOS, Cmd-Left and Cmd-Right are supposed to be basically the same as "Home" and "End."

However, in QtConsole, Cmd-Left has a strange behavior: it literally brings you to the beginning of the line - aka before the prompt. Then, since the cursor is positioned at an "invalid" location, once you type anything it brings you right back to the end of the line.

Adding 8 lines of code to `_event_filter_console_keypress(self, event)` in `console-widget.py` yields the correct behavior:

```
...
    def _event_filter_console_keypress(self, event):
        """ Filter key events for the underlying text widget to create a
            console-like interface.
        """
        intercepted = False
        cursor = self._control.textCursor()
        position = cursor.position()
        key = event.key()
        ctrl_down = self._control_key_down(event.modifiers())
        alt_down = event.modifiers() & QtCore.Qt.AltModifier
        shift_down = event.modifiers() & QtCore.Qt.ShiftModifier

#### BEGIN INSERT
        cmd_down = \
            sys.platform == "darwin" and \
            self._control_key_down(event.modifiers(), include_command=True)
        if cmd_down:
            if key == QtCore.Qt.Key_Left:
                key = QtCore.Qt.Key_Home
            elif key == QtCore.Qt.Key_Right:
                key = QtCore.Qt.Key_End
#### END INSERT
        #------ Special modifier logic -----------------------------------------
...
```

You could also probably make a `_cmd_down_remap` similar to the `_ctrl_down_remap` that you already have, but this seemed simpler.

### LINKED ISSUE #484: Pressing Shift causes window to scroll back to current prompt if scrolled up
If you are trying to scroll up to copy/paste some text from further up, you might do this:

1. Scroll up to where you want the selection to begin
2. Click to put the cursor at the selection begin point
3. Hold the "shift" button
4. Scroll to where you want the selection to end
5. Click to put the cursor at the selection end point

The problem is that when you type "shift", qtconsole causes the text cursor to jump back to the prompt, so you can't copy or paste anything. This happens with quite a few keys - shift, option, etc - but not ctrl.

It's because of this in `console_widget.py`, at the very end of `_event_filter_console_keypress(self, event)`

```
        if not (self._control_key_down(event.modifiers(), include_command=True)
                or key in (QtCore.Qt.Key_PageUp, QtCore.Qt.Key_PageDown)
                or (self._executing and not self._reading)):
            self._keep_cursor_in_buffer()
```

That call to `self._keep_cursor_in_buffer()` is what causes it to reset the position.

There are quite a few modifiers that cause this - not just shift but literally any key except for the three exceptions above - page up, page down, and control. It seems to me that the cleanest way to do this is to simply add the following line

```
        if not (self._control_key_down(event.modifiers(), include_command=True)
                or key in (QtCore.Qt.Key_PageUp, QtCore.Qt.Key_PageDown)
                or (self._executing and not self._reading)
                or (event.text() == "" and not key in (QtCore.Qt.Key_Up, QtCore.Qt.Key_Down))): ## <-- ADDED!
            self._keep_cursor_in_buffer()

```

This only scrolls back to the bottom only if the user actually typed something which has text in it, like a character, rather than just typing some modifier or special key (e.g. shift, page up/page down, maybe print screen), with the one exception of if they've typed the up or down keys (which should scroll through the history). Then everything seems to work the way you'd expect.

battaglia01:
These implement the keyboard improvements from #483 and #484. 

Fixes #483
Fixes #484

### DIFF
--- a/qtconsole/console_widget.py
+++ b/qtconsole/console_widget.py
@@ -1228,9 +1228,10 @@ class ConsoleWidget(MetaQObjectHasTraits('NewBase', (LoggingConfigurable, superQ
         alt_down = event.modifiers() & QtCore.Qt.AltModifier
         shift_down = event.modifiers() & QtCore.Qt.ShiftModifier
 
-        cmd_down = \
-            sys.platform == "darwin" and \
+        cmd_down = (
+            sys.platform == "darwin" and
             self._control_key_down(event.modifiers(), include_command=True)
+        )
         if cmd_down:
             if key == QtCore.Qt.Key_Left:
                 key = QtCore.Qt.Key_Home
--- a/qtconsole/console_widget.py
+++ b/qtconsole/console_widget.py
@@ -1228,6 +1228,20 @@ class ConsoleWidget(MetaQObjectHasTraits('NewBase', (LoggingConfigurable, superQ
         alt_down = event.modifiers() & QtCore.Qt.AltModifier
         shift_down = event.modifiers() & QtCore.Qt.ShiftModifier
 
+        cmd_down = \
+            sys.platform == "darwin" and \
+            self._control_key_down(event.modifiers(), include_command=True)
+        if cmd_down:
+            if key == QtCore.Qt.Key_Left:
+                key = QtCore.Qt.Key_Home
+            elif key == QtCore.Qt.Key_Right:
+                key = QtCore.Qt.Key_End
+            elif key == QtCore.Qt.Key_Up:
+                ctrl_down = True
+                key = QtCore.Qt.Key_Home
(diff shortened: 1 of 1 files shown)


ccordoba12 (commented):
Hey @battaglia01, thanks a lot for your contribution! I left a minor suggestion for you, otherwise looks good to me.

ccordoba12 (approved):
Thanks @battaglia01!

PR #491, opened 2021-07-12
Fix DeprecationWarning: invalid escape sequence.

Carreau:
Replace \, by \\, which is how python currently interprets it.
r"" would be incorrect as the \x1b would be escaped.

Checked that ANSI_PATTERN is unchanged to:

In [7]: ANSI_PATTERN
Out[7]: '\x01?\x1b(\\[(.*?)([ABCDEFGHJKSTfmnsu])|\\](.*?)[\x07\x1b])\x02?'


ccordoba12 (approved):
Thanks @Carreau! Looks good to me.

PR #493, opened 2021-07-30
Test jupyter_client master

davidbrochart:


davidbrochart:
Trigger CI.

davidbrochart:
Trigger CI.

davidbrochart:
The errors on Mac and Windows seem related to kernel provisioning:
```
        if not KPF.instance(parent=self.parent).is_provisioner_available(kspec):
>           raise NoSuchKernel(kernel_name)
E           jupyter_client.kernelspec.NoSuchKernel: No such kernel named python3
```
Maybe @kevin-bates has some ideas?

kevin-bates:
Sorry for not seeing this sooner David.  

This is a bit perplexing because this particular message about a missing `local-provisioner` is essentially an invariant condition since the `local-provisioner` is packaged with `jupyter_client`.  This implies there's some snafu with the registration of the [entrypoint](https://github.com/jupyter/jupyter_client/blob/279de8962eb00db4204146cfb56e74f9331c2972/setup.py#L73-L75) since this message is produced whenever an implied (as in this case) or referenced (as when the `kernel.json` actually has a `kernel_provsioner` stanza) provisioner cannot be loaded and is, therefore, deemed unavailable.

In looking into this closer and, in particular, the `pip install` statement [the CI is doing](https://github.com/jupyter/qtconsole/pull/493/files#diff-6a7baa40c387bd464cf28f909290c10a1ec71f4e8129401f2127cdddd9b40c99R37), it looks like the previous dist-info directory relative to 6.1.12 still exists and alongside the dist-info corresponding to the 7.0 release.  Since this is where the entrypoint definitions reside, it must be the case that the entrypoint machinery just locates the first dist-info directory it encounters - thereby treating the `local-provisioner` as non-existent.

If I preceed the CI pip install with `pip uninstall -y jupyter_client` or replace `--ignore-installed` with `--force-reinstall` only one dist-info directory exists and contains the appropriate entrypoint references.

Once cleanly installed, one should see these log statements after invoking `jupyter qtconsole --debug`:
```
[JupyterQtConsoleApp] Instantiating kernel 'Python 3' with kernel provisioner: local-provisioner
[JupyterQtConsoleApp] Starting kernel: ['/opt/anaconda3/envs/jupyter-server-dev/bin/python', '-m', 'ipykernel_launcher', '-f', '/Users/kbates/Library/Jupyter/runtime/kernel-9598.json']
```

I also suspect that this issue exists on all the platforms and that, on Linux, some other issue is happening prior to this - but I haven't looked into that at all.

.github/workflows/linux-tests.yml
```diff
@@ -37,7 +37,9 @@ jobs:
            python-version: ${{ matrix.PYTHON_VERSION }}
       - name: Install dependencies
         shell: bash -l {0}
-        run: conda env update --file requirements/environment.yml
+        run: |
+          conda env update --file requirements/environment.yml
+          pip install https://github.com/jupyter/jupyter_client/archive/master.zip --ignore-installed
```
kevin-bates, line 42: Per https://github.com/jupyter/qtconsole/pull/493#issuecomment-893112222:
```suggestion
          pip install https://github.com/jupyter/jupyter_client/archive/master.zip --force-reinstall
```
and ditto for the others.

davidbrochart:
Thanks Kevin, `--force-reinstall` solved the issue. Now we have these errors on Windows:
- `Windows fatal exception: code 0xc0000139`
- an apparently empty connection file.

kevin-bates:
Hi David, I suspect the two Windows issues are related since the fatal exception is occurring within the Windows-only portion of attempting to write the connection file in a secure manner.  Because that exception occurs, there's no connection file.   Aside from starting the kernel process (as some `Popen` options are different on Windows), this is the only location that would be platform-specific (practically speaking).

In quickly googling the fatal exception, I have found a couple of instances (one of which appears to be in Chinese - and may just be a translation of [this page](https://stackoverflow.com/questions/53199675/importing-tensorflow-makes-python-3-6-5-error)) suggesting that `qt` and `pyqt` packages be updated.  I don't know if this is reaching too far, but I wonder if this could be another installation snafu, perhaps introduced by `--force-reinstall` that warrants `pyqt5` being updated following the reinstall of `jupyter_client`?

It might also be a good idea to include `pywin32` in any "post-install" updates since that's where the issue is being encountered.

FWIW, this [PR](https://github.com/jupyter/jupyter_core/pull/230) in core removes the `pywin32` dependency, but this currently resides in only core's master branch.

davidbrochart:
Thanks Kevin for investigating, I think this is getting out of our original scope of checking for any issue with jupyter_client v7.0. I consider we have done enough, and maybe the Windows issue is for the qtconsole team to look at. But feel free to comment @ccordoba12 if you disagree.

.github/workflows/windows-tests.yml
```diff
@@ -32,7 +32,9 @@ jobs:
            python-version: ${{ matrix.PYTHON_VERSION }}
       - name: Install package dependencies
         shell: bash -l {0}
-        run: conda env update --file requirements/environment.yml
+        run: |
+          conda env update --file requirements/environment.yml
+          pip install https://github.com/jupyter/jupyter_client/archive/master.zip --force-reinstall
```
kevin-bates, line 37: In digging into this a bit, I found that the conda version of `pywin32` was `228` while the version added during the reinstall was `301`.  I found that if I apply the following change, the [Windows tests start passing](https://github.com/kevin-bates/qtconsole/actions/runs/1102574719):
```suggestion
          pip install https://github.com/jupyter/jupyter_client/archive/master.zip --force-reinstall
          pip install --upgrade pywin32==228
```
I don't know if there's some incompatibility with other "pieces" installed via conda then whacked by the reinstall, but it seems like reverting the `pywin32` version to what conda laid down does some good.

My packaging-fu isn't sufficient to troubleshoot this further, but this might allow this PR to move forward.

ccordoba12: Are there new dependencies in jupyter-client 7? If not, I suggest to install it from master with `--no-deps`. If yes, is it possible to install the new dependencies with conda?

There shouldn't be problems by reinstalling pywin32 with pip, as @kevin-bates suggests, because pywin32 doesn't have dependencies (if I recall correctly). But better to avoid it if possible.

kevin-bates: 7.0 adds a dependency on entrypoints.

kevin-bates: I'll try adding `conda install entrypoints`

ccordoba12: Ok, since that's present in Anaconda, please run

    conda install entrypoints

before the `pip install` command that installs jupyter-client from master, then pass `--no-deps` to it. That should solve the problem.

kevin-bates: Thanks @ccordoba12 - that indeed is a better approach.  I had to also add `nest-asycio` as well: https://github.com/kevin-bates/qtconsole/actions/runs/1103052025

I'll update the within the review.

.github/workflows/linux-tests.yml
```diff
@@ -37,7 +37,9 @@ jobs:
            python-version: ${{ matrix.PYTHON_VERSION }}
       - name: Install dependencies
         shell: bash -l {0}
-        run: conda env update --file requirements/environment.yml
+        run: |
+          conda env update --file requirements/environment.yml
+          pip install https://github.com/jupyter/jupyter_client/archive/master.zip --force-reinstall
```
kevin-bates, line 42: Per this conversation (https://github.com/jupyter/qtconsole/pull/493#discussion_r683777815) the recommended approach would be the following across the three platforms:
```suggestion
          conda install entrypoints nest-asyncio
          pip install https://github.com/jupyter/jupyter_client/archive/master.zip --no-deps
```

davidbrochart:
@kevin-bates @ccordoba12 thanks, that worked!
I don't think we want to merge this PR as-is, as it's testing jupyter_client master. But it can be useful to have it test a jupyter_client release, as it should pick up pre-releases.

davidbrochart:
Actually I think it was working in the previous commit because jupyter_client 6 was picked up. I added the following to make sure:
```bash
python -c 'import jupyter_client; print("jupyter_client", jupyter_client.__version__)'
```
and now I can see that jupyter_client 7 is picked up only if I add `--ignore-installed` to the `pip` command.

meeseeksmachine:
This pull request has been mentioned on **Jupyter Community Forum**. There might be relevant details there:

https://discourse.jupyter.org/t/kernel-python-3-is-referencing-a-kernel-provisioner-local-provisioner-that-is-not-available-ensure-the-appropriate-package-has-been-installed-and-retry/10436/2

ccordoba12:
Closing because this was only for testing.

Issue #350, opened 2019-07-07
ANSI escape sequences other than color


mdietrich16:
I am using the jupyter qtconsole through spyder, but I have tested the following in a standalone console aswell.

I know that certain ANSI escape sequences are being parsed, like simple colors and TrueColors, but is it possible to use stuff like _'\027[A'_ (cursor up) or implement it myself?

I'm not very profound in stuff like this, but if I could get a little nudge in the right direction I might be able to at least help myself if it is not overly complicated and hard to pull off.

Oh, and I saw that line 27 in ansi_code_processor.py actually says
`# FIXME: Not implemented in AnsiCodeProcessor.`

So do you just need people or good ideas, because I would happy to fix the first :)

ccordoba12:
We need people, so you're more than welcome to help us to improve this.

wmvanvliet:
PRs are always welcome!

PropagandaP4nd4:
Well, seeing that I have never worked with the intrinsics of this project, I would need some pointer as to what and where to look to _try_ to improve this, as I said. But other than the first steps, happy to help!

sylvain-bougnoux:
Yes, same for erase line, eg:
`print('Hello','\x1b[2K\rBye')`
Should print 'Bye' and not 'Byelo'.
It does work on a windows terminal (both with python & ipython).

ma-sadeghi:
Can I ask what's the status of this issue? Is it hard to fix?

ccordoba12:
It shouldn't be too hard to fix, we just haven't had time to sit down and do it.

ma-sadeghi:
Could you please give me some pointers so I can take a crack at it? I don't know where to start.

ccordoba12:
Our ANSI processor is implemented here:

https://github.com/jupyter/qtconsole/blob/master/qtconsole/ansi_code_processor.py

PR #499, opened 2021-09-14
Fix ANSI code "Erase in Line"

sylvain-bougnoux:
It is related to #350.
It now fixes the K ANSI code (Erase in Line) supporting:

-  0K for 'until end', 
- 1K for 'until start', 
- 2K for all line (whole). 

1/ I hope my code does not break anything; it can, as I had to change '\r' carriage return & '\b' that strangely used to keep the cursor in position.  Somehow I have implemented a more natural 'replace text' mode (instead than default insert mode).

2/ There was a misunderstanding of the specs requiring not to move the cursor ([ANSI code](https://en.wikipedia.org/wiki/ANSI_escape_code#CSI_(Control_Sequence_Introducer)_sequences)).
Therefore I had also to replace sometimes the removed text by spaces.

I guess it is WIP.

Some examples, giving similar results in Python & IPython, AFAI've checked.
```
print('Hello\x1b[1KBye')    # "     Bye"
print('Hello\x1b[0KBye')    # "HelloBye"
print('Hello\r\x1b[0KBye')  # "Bye"
print('Hello\r\x1b[1KBye')  # "Byelo"
print('Hello\r\x1b[2KBye')  # "Bye"
print('Hello\x1b[2K\rBye')  # "Bye"
```

### DIFF
--- a/qtconsole/ansi_code_processor.py
+++ b/qtconsole/ansi_code_processor.py
@@ -144,7 +144,6 @@ class AnsiCodeProcessor(object):
         substring = SPECIAL_PATTERN.sub(self._replace_special, raw)
         if substring or self.actions:
             yield substring
-            self.actions = []
 
         if last_char is not None:
             self.actions.append(NewLineAction('newline'))
--- a/qtconsole/ansi_code_processor.py
+++ b/qtconsole/ansi_code_processor.py
@@ -144,6 +144,7 @@ class AnsiCodeProcessor(object):
         substring = SPECIAL_PATTERN.sub(self._replace_special, raw)
         if substring or self.actions:
             yield substring
+            self.actions = []
 
         if last_char is not None:
             self.actions.append(NewLineAction('newline'))
--- a/qtconsole/console_widget.py
+++ b/qtconsole/console_widget.py
@@ -2181,11 +2181,13 @@ class ConsoleWidget(MetaQObjectHasTraits('NewBase', (LoggingConfigurable, superQ
                 if substring is not None:
                     format = self._ansi_processor.get_format()
                     pos = cursor.position()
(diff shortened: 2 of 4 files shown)


sylvain-bougnoux:
Only 'test_prompt_cursors' keeps failing (just the last assert trying to insert before prompt). It is normal as the default is now to replace, instead than to insert. Either I fix it to make the pytest pass (but I am afraid it won't be natural), either one change the test (because somehow the failure is normal)? However it is difficult to validate that this change of default is harmless, maybe there is not enough test?
I am definitely not familiar enough with the code to judge.
Somehow from a design point of view, it would be nice to distinguish the insert from the replace mode (as does the keyboard key 'inser').

ccordoba12 (approved):
Looks good to me, thanks @sylvain-bougnoux!

sylvain-bougnoux:
@ccordoba12 thanks for your trust Carlos, but I've just updated and I see strange artifact in the console. I passed the CI tests, but I am afraid the tests are weak as I suggested? May it is another issue, maybe it is my fault?

In the qtconsole (I test on spyder) type ('>' refers to the prompt "In [...]")
`>text='test'`
now type`>text` and press return at the end. It works as expected.
but type `>text`, press left key, press return. 'Out [...]' does not appear and the result is appended at the end instead...

ccordoba12:
Ok, do you have time to take a look at that problem these days?

sylvain-bougnoux:
@ccordoba12 I had to do a new PR (see #506). I did a fast fix. My solution is simple but maybe could have been formulated differently. Actually I was running out of time. I also have not added a CI-test for this case.

ccordoba12:
Ok, no problem and thanks for your help with that!

PR #505, opened 2021-11-06
Fix hidden execution requests

impact27:
The requests are not necessarily sent after receiving the reply from the previous request. If several request are sent in rapid succession, `self._hidden` becomes meaningless. The hidden state of each request must therefore be saved.

Also, the prompt was sometimes requested several times, so to avoid multiple confusing prompt, a check is added.

### DIFF
--- a/qtconsole/base_frontend_mixin.py
+++ b/qtconsole/base_frontend_mixin.py
@@ -144,7 +144,9 @@ class BaseFrontendMixin(object):
 
     def include_output(self, msg):
         """Return whether we should include a given output message"""
-        if self._hidden:
+        msg_id = msg['parent_header']['msg_id']
+        info = self._request_info['execute'].get(msg_id)
+        if info and info.hidden:
             return False
         from_here = self.from_here(msg)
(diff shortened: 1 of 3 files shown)


qtconsole/base_frontend_mixin.py
```diff
@@ -144,8 +144,12 @@ def from_here(self, msg):
 
     def include_output(self, msg):
         """Return whether we should include a given output message"""
-        if self._hidden:
-            return False
+        if msg['parent_header']:
+            # If parent message is from hidden execution, do not include
```
ccordoba12, line 148:
```suggestion
            # If parent message is from hidden execution, don't include it.
```

Small grammar improvement to this comment.

ccordoba12 (commented):
Thanks @impact27 for your work on this! I left a tiny suggestion, otherwise looks good to me.

impact27:
Ok ready, I fixed a small issue with 'silent_exec_callback'

ccordoba12 (approved):
Thanks @impact27 for your help with this!

PR #509, opened 2021-11-16
Fix error when deleting CallTipWidget

### LINKED ISSUE #507: Attempt to delete CallTipWidget after the C/C++ object has been deleted
When shutting down an [application](https://nexpy.github.io/nexpy/) which has an embedded QtConsole, I get the following error with v5.2.0, which does not occur in v5.1.1. 
```
RuntimeError                              Traceback (most recent call last)
~/opt/miniconda3/envs/py38/lib/python3.8/site-packages/qtconsole/call_tip_widget.py in <lambda>()
     21         assert isinstance(text_edit, (QtWidgets.QTextEdit, QtWidgets.QPlainTextEdit))
     22         super().__init__(None, QtCore.Qt.ToolTip)
---> 23         text_edit.destroyed.connect(lambda: self.deleteLater())
     24 
     25         self._hide_timer = QtCore.QBasicTimer()

RuntimeError: wrapped C/C++ object of type CallTipWidget has been deleted
```
It appears to happen when calling the QApplication `quit` function, so I guess the standard PyQt shutdown process doesn't guarantee that the C/C++ ToolTip widget is destroyed after the QTextEdit widget. I can get around this by explicitly closing the console widget, which is a sub-class of FrontendWidget, before calling `quit`.

I am raising this as an issue because I wanted to find out if this was the expected behavior. Are you assuming that the FrontendWidget is always explicitly closed when shutting down an application using QtConsole? That wasn't necessary before the change. It might be possible to prevent this by subclassing the CallTipWidget `deleteLater` function to catch this exception, which would ensure that other applications don't break.

rayosborn:
* Subclasses the CallTipWidget `deleteLater` function to catch errors caused by the prior deletion of the underlying C/C++ widget.
* Supersedes the exception handling in the CallTipWidget `__init__` function.
* Simplifies the specification of the slot connected to the `destroyed` signal.
* Fixes #507.

### DIFF
--- a/qtconsole/call_tip_widget.py
+++ b/qtconsole/call_tip_widget.py
@@ -21,13 +21,7 @@ class CallTipWidget(QtWidgets.QLabel):
         assert isinstance(text_edit, (QtWidgets.QTextEdit, QtWidgets.QPlainTextEdit))
         super().__init__(None, QtCore.Qt.ToolTip)
 
-        # This necessary to avoid an error when the widget was already
-        # deleted.
-        # Fixes jupyter/qtconsole#507
-        try:
-            text_edit.destroyed.connect(lambda: self.deleteLater())
-        except RuntimeError:
-            pass
+        text_edit.destroyed.connect(self.deleteLater)
 
         self._hide_timer = QtCore.QBasicTimer()
         self._text_edit = text_edit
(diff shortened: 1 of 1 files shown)


ccordoba12 (approved):
Looks good to me, thanks @rayosborn!

PR #524, opened 2022-02-22
PR: Add testing jobs using pip to use latest PyQt5 version available 

dalthviz:
Related to https://github.com/spyder-ide/spyder/pull/16409


.github/workflows/linux-tests.yml
```diff
@@ -21,31 +21,44 @@ jobs:
...
            python-version: ${{ matrix.PYTHON_VERSION }}
-      - name: Install dependencies
+      - name: Install dependencies with conda
+        if: matrix.INSTALL_TYPE == 'conda'
         shell: bash -l {0}
         run: conda env update --file requirements/environment.yml
+      - name: Install dependencies with pip
+        if: matrix.INSTALL_TYPE == 'pip'
+        shell: bash -l {0}
+        run: pip install -r requirements/requirements.txt
```
ccordoba12, line 52:
```suggestion
        run: |
          pip install -e .[test]
          pip install PyQt5 coveralls pytest-cov
```

This will avoid declaring an extra `requirements.txt` file and will also test that our `setup.py` is working as expected.

ccordoba12 (commented):
Thanks @dalthviz for your help with this! I left a small suggestion for you, otherwise looks good to me.

PR #525, opened 2022-03-05
add PyQt6 support

eyllanesc:


svartkanin:
Any chance this might get merged soonish?

qtconsole/call_tip_widget.py
```diff
@@ -5,6 +5,8 @@
 # System library imports
 from qtpy import QtCore, QtGui, QtWidgets
 
+from .util import IsQt6
```
ccordoba12, line 8: Please rename `IsQt6` to `IS_QT6`. That's because we don't use camel-case names for constants.

qtconsole/console_widget.py
```diff
@@ -15,16 +15,20 @@
 
 from qtpy import QtCore, QtGui, QtPrintSupport, QtWidgets
 
-from traitlets.config.configurable import LoggingConfigurable
 from qtconsole.rich_text import HtmlExporter
 from qtconsole.util import MetaQObjectHasTraits, get_font, superQ
+
 from ipython_genutils.text import columnize
+
```
ccordoba12, line 22: Please remove this blank here.

qtconsole/console_widget.py
```diff
@@ -432,7 +436,7 @@ def eventFilter(self, obj, event):
         # Make middle-click paste safe.
         elif getattr(event, 'button', False) and \
                 etype == QtCore.QEvent.MouseButtonRelease and \
-                event.button() == QtCore.Qt.MidButton and \
+                event.button() == (QtCore.Qt.MouseButton.MiddleButton if IsQt6 else QtCore.Qt.MidButton) and \
```
ccordoba12, line 439: Could you remove the continuation lines here and instead use parenthesis to wrap the code in this `elif`?

I'm referring to change this to

```python
elif (
    getattr(event, 'button', False) and
    etype == QtCore.QEvent.MouseButtonRelease and
    ...
):

File: qtconsole/tests/test_00_console_widget.py [hunk 4]
```diff
@@ -2,12 +2,17 @@
 import sys
 
 from flaky import flaky
+
```
ccordoba12, line 5: Please remove this blank.

qtconsole/tests/test_00_console_widget.py
```diff
@@ -2,12 +2,17 @@
...
 import pytest
+
+from qtconsole.util import IsQt6
```
ccordoba12, line 8: This line in unnecessary because the same import is below. Please remove it.

qtconsole/tests/test_jupyter_widget.py
```diff
@@ -2,12 +2,15 @@
 import sys
 
 import pytest
+
```
ccordoba12, line 5: Please remove this blank.

qtconsole/tests/test_jupyter_widget.py
```diff
@@ -70,7 +73,18 @@ def test_other_output(self):
         ))
 
         # Check proper syntax highlighting
-        self.assertEqual(document.toHtml(), (
+        html = (
```
ccordoba12, line 76: Please use

```python
if IS_QT6:
    html = ...
else:
    html = ...
```

instead of trying to pack everything in an inline if.

qtconsole/tests/test_jupyter_widget.py
```diff
@@ -81,7 +95,8 @@ def test_other_output(self):
             '<p style=" margin-top:0px; margin-bottom:0px; margin-left:0px; margin-right:0px; -qt-block-indent:0; text-indent:0px;"><span style=" color:#000080;">\xa0\xa0\xa0\xa0\xa0\xa0\xa0\xa0\xa0\xa0\xa0...:</span> b = range(10)</p>\n'
             '<p style="-qt-paragraph-type:empty; margin-top:0px; margin-bottom:0px; margin-left:0px; margin-right:0px; -qt-block-indent:0; text-indent:0px;"><br /></p>\n'
             '<p style=" margin-top:0px; margin-bottom:0px; margin-left:0px; margin-right:0px; -qt-block-indent:0; text-indent:0px;"><span style=" color:#000080;">In [</span><span style=" font-weight:600; color:#000080;">2</span><span style=" color:#000080;">]:</span> </p></body></html>'
-        ))
+        )
```
ccordoba12, line 98: Please add a blank line separating the if/else I mentioned and the assert below.

ccordoba12 (commented):
Thanks @eyllanesc for your help with this! It looks pretty good!

I left a small review for you, mostly about style issues.

eyllanesc:
@ccordoba12 Thanks for the comments, I will solve them soon. Do you use any tool to format code? for example black.

qtconsole/tests/test_jupyter_widget.py
```diff
@@ -70,7 +72,21 @@ def test_other_output(self):
         ))
 
         # Check proper syntax highlighting
-        self.assertEqual(document.toHtml(), (
+        if IS_QT6:
```
ccordoba12, line 75: You missed this comment from my previous review:

Please use

```python
if IS_QT6:
    html = ...
else:
    html = ...
```

instead of trying to pack everything in an inline if.

ccordoba12:
> Do you use any tool to format code(style)? for example [black](https://github.com/psf/black).

No, we don't. And we prefer not to because manual adjustments are sometimes better.

qtconsole/tests/test_jupyter_widget.py
eyllanesc: If I put something like: 

```python
if IS_QT6:
    html = '<!DOCTYPE HTML PUBLIC "-//W3C//DTD HTML 4.0//EN" "http://www.w3.org/TR/REC-html40/strict.dtd">\n'
    '<html><head><meta name="qrichtext" content="1" /><meta charset="utf-8" /><style type="text/css">\n'
    'p, li { white-space: pre-wrap; }\n'
else:
    html = '<!DOCTYPE HTML PUBLIC "-//W3C//DTD HTML 4.0//EN" "http://www.w3.org/TR/REC-html40/strict.dtd">\n'
    '<html><head><meta name="qrichtext" content="1" /><style type="text/css">\n'
    'p, li { white-space: pre-wrap; }\n'
```

So html only takes the value of the first line and not all three lines. 

If you wanted to keep that format you would have to add `\` to the end of each line: 



```python
if IS_QT6:
    html = '<!DOCTYPE HTML PUBLIC "-//W3C//DTD HTML 4.0//EN" "http://www.w3.org/TR/REC-html40/strict.dtd">\n' \
    '<html><head><meta name="qrichtext" content="1" /><meta charset="utf-8" /><style type="text/css">\n' \
    'p, li { white-space: pre-wrap; }\n'
else:
    html = '<!DOCTYPE HTML PUBLIC "-//W3C//DTD HTML 4.0//EN" "http://www.w3.org/TR/REC-html40/strict.dtd">\n' \
    '<html><head><meta name="qrichtext" content="1" /><style type="text/css">\n' \
    'p, li { white-space: pre-wrap; }\n'
```

Is that what you want? I prefer to keep using `()`. So in the end it is your decision, which option do you want?

ccordoba12: > If you wanted to keep that format you would have to add \ to the end of each line:

Nop, you can also enclose the strings in parenthesis:

```python
if IS_QT6:
    html = (
        '<!DOCTYPE HTML PUBLIC "-//W3C//DTD HTML 4.0//EN" "http://www.w3.org/TR/REC-html40/strict.dtd">\n'
        '<html><head><meta name="qrichtext" content="1" /><meta charset="utf-8" /><style type="text/css">\n'
        'p, li { white-space: pre-wrap; }\n'
    )
else:
    html = (
        '<!DOCTYPE HTML PUBLIC "-//W3C//DTD HTML 4.0//EN" "http://www.w3.org/TR/REC-html40/strict.dtd">\n'
        '<html><head><meta name="qrichtext" content="1" /><style type="text/css">\n'
        'p, li { white-space: pre-wrap; }\n'
    )
```

That avoids the continuation lines and gives the result I'd like to see.

eyllanesc:
@ccordoba12 See the changes.

.github/workflows/linux-tests.yml
```diff
@@ -22,21 +22,19 @@ jobs:
...
-        if: matrix.INSTALL_TYPE == 'pip'
+        uses: actions/checkout@v3
+      - name: Install System Packages
         run: |
           sudo apt-get update
-          sudo apt-get install -qq pyqt5-dev-tools
+          sudo apt-get install -y --no-install-recommends '^libxcb.*-dev' libx11-xcb-dev libglu1-mesa-dev libxrender-dev libxi-dev libxkbcommon-dev libxkbcommon-x11-dev
       - name: Install Conda
-        uses: conda-incubator/setup-miniconda@v2
+        uses: conda-incubator/setup-miniconda@v2.1.1
```
ccordoba12, line 37: Why ping to this specific version? Shouldn't be better to use `v2` to bring the latest 2.x.x version of this action?

qtconsole/console_widget.py
```diff
@@ -411,6 +414,7 @@ def eventFilter(self, obj, event):
             text widgets.
         """
         etype = event.type()
+        middle_button = QtCore.Qt.MouseButton.MiddleButton if IS_QT6 else QtCore.Qt.MidButton
```
ccordoba12, line 417: What do you think about adding this alias to QtPy instead?

@dalthviz, what do you think? (@dalthviz is the maintainer of QtPy).

qtconsole/tests/test_00_console_widget.py
```diff
@@ -274,11 +277,10 @@ def test_erase_in_line(self):
...
     def test_link_handling(self):
-        noKeys = QtCore.Qt
-        noButton = QtCore.Qt.MouseButton(0)
-        noButtons = QtCore.Qt.MouseButtons(0)
-        noModifiers = QtCore.Qt.KeyboardModifiers(0)
-        MouseMove = QtCore.QEvent.MouseMove
+        noButton = QtCore.Qt.MouseButton.NoButton if IS_QT6 else QtCore.Qt.MouseButton(0)
+        noButtons = QtCore.Qt.MouseButton.NoButton if IS_QT6 else QtCore.Qt.MouseButtons(0)
+        noModifiers = QtCore.Qt.KeyboardModifier.NoModifier if IS_QT6 else QtCore.Qt.KeyboardModifiers(0)
+        MouseMove = QtCore.QEvent.Type.MouseMove if IS_QT6 else QtCore.QEvent.MouseMove
```
ccordoba12, line 283: The same about this enum? What if we add it to QtPy instead?

@dalthviz?

ccordoba12 (commented):
Thanks for continued work on this @eyllanesc! A much simpler review for you now.

qtconsole/console_widget.py
dalthviz: Checking this with QtPy 2.0.1 seems like the `QtCore.Qt.MidButton` alias for `QtCore.Qt.MouseButton.MiddleButton` is not available with PyQt6 (looking at the code seems like we added the missing mapping for PySide6 but not for PyQt6) however you can use `QtCore.Qt.MiddleButton`  with Qt 5 and 6 (PyQt5/6 and PySide2/6). We probably should add the missing alias for PyQt6 but for the moment changing this line to something like the following should be enought:

```python
        middle_button = QtCore.Qt.MiddleButton
```

qtconsole/tests/test_00_console_widget.py
dalthviz: Checking this I was able to use the scoped and unscoped versions without issue with QtPy 2.0.1 with PyQt5/6 and PySide5/6:

```python
QtCore.Qt.MouseButton.NoButton -> QtCore.Qt.NoButton
QtCore.Qt.KeyboardModifier.NoModifier -> QtCore.Qt.NoModifier
QtCore.QEvent.Type.MouseMove -> QtCore.QEvent.MouseMove
```

qtconsole/console_widget.py
dalthviz: I think if you use the default channel you will get QtPy 1.11.2 if you use the conda-forge channel you should get the latest version (2.0.1). Could you try to use the conda-forge channel when installing?

eyllanesc: @dalthviz Thanks for the comments.

I'm a newbie to conda, could you point out what change I would have to make to linux-tests.yml: https://github.com/eyllanesc/qtconsole/blob/ce1373fb83288b8227be4fa1425db5d479727801/.github/workflows/linux-tests.yml to use the conda-forge channel?

dalthviz: Glad to help @eyllanesc For the test workflow I think the `setup-miniconda@v2` action has options to set the channels and priority. Something like this in the `with:` section of the step should help with this:

```
channels: conda-forge,defaults
channel-priority: strict
```

eyllanesc: @dalthviz It did not help many unfortunately, if I use `channel-priority: strict` then the installation process fails: https://github.com/eyllanesc/qtconsole/runs/5483917012?check_suite_focus=true, if I remove it it installs qtpy 1.11.2: https://github.com/eyllanesc/qtconsole/actions/runs/1958675144

dalthviz: I think I found the thing here, the channels definition should be done in the `requirements/environment.yml` file itself:

https://github.com/eyllanesc/qtconsole/blob/ce1373fb83288b8227be4fa1425db5d479727801/requirements/environment.yml#L1-L3

So my guess is that you will need to add there the conda-forge channel as first item in the `channels` entry

dalthviz: Maybe you could add a constraint in the `environment.yml` for pyqt. Something like `- pyqt >= 5.12`

eyllanesc: @dalthviz Great, I already corrected other things and now if you install more recent versions of the libraries. Later I will remove redundant code regarding the most recent version of QtPy.

.github/workflows/linux-tests.yml
```diff
@@ -22,25 +23,25 @@ jobs:
...
           sudo apt-get update
-          sudo apt-get install -qq pyqt5-dev-tools
+          sudo apt-get install -y --no-install-recommends '^libxcb.*-dev' libx11-xcb-dev libglu1-mesa-dev libxrender-dev libxi-dev libxkbcommon-dev libxkbcommon-x11-dev
       - name: Install Conda
         uses: conda-incubator/setup-miniconda@v2
         with:
            activate-environment: test
            auto-update-conda: false
            auto-activate-base: false
+           channels: conda-forge,defaults,spyder-ide
```
ccordoba12, line 43: Let's remove the `defaults` and `spyder-ide` channels from here because they're not necessary.

requirements/environment.yml
```diff
@@ -1,4 +1,5 @@
 channels:
+  - conda-forge
   - defaults
   - spyder-ide
```
ccordoba12, line 4: Let's remove these two channels from here too. Since all packages required for testing are part of conda-forge, they are not necessary anymore.

ccordoba12 (commented):
Thanks @eyllanesc for your latest changes and @dalthviz for chiming in!

Last two comments, then this should be ready.

ccordoba12 (approved):
Thanks for your help @eyllanesc! Great work here.

PR #527, opened 2022-03-10
Run tests with Python 3.10

eyllanesc:
qtconsole supports python 3.10 as shown in the setup.py:

https://github.com/jupyter/qtconsole/blob/cbffab2e0368fa42a6e10d0427c74aceaa1ee3d3/setup.py#L99

But the GitHub workflow doesn't cover them. This PR tries to solve that problem.


ccordoba12 (commented):
Thanks @eyllanesc! Looks good to me.

Please squash your two commits and leave a single one because both have the same commit message. Afterwards, I'll merge your PR.

PR #530, opened 2022-03-29
Remove scripts that are duplicate to entry points in setup.py

mgorny:
Remove the explicit "jupyter-qtconsole" script that is also covered
by an entry point.  Having both is redundant and causes "installer"
to fail on the resulting wheel.  Furthermore, installing explicit
scripts is deprecated in setuptools in favor of entry points.

Fixes #529

PR #532, opened 2022-04-03
Empty commit to rerun CI

mgorny:
To test if CI works correctly on master, as requested in #530.


ccordoba12:
Thanks for doing this @mgorny for your help with this. This problem was probably introduced by IPykernel 6.11, which we'll try to fix in the next weeks.

mgorny:
Should I file a PR to pin it to earlier version for now?

PR #536, opened 2022-05-03
Replace _get_block_plain_text() with QTextBlock.text() to fix segfault on Qt 6

firai:
Replace `_get_block_plain_text()` with `QTextBlock.text()`.

Fixes #531.

PR #538, opened 2022-05-04
Fix missing argument error when switching syntax highlighting styles on PySide2 & PySide6

### LINKED ISSUE #537: Changing syntax highlighting style doesn't work with PySide2 & PySide6
If I try to change the syntax highlighting style when PySide2 or PySide6 is installed, I get the following error. However, changing the style seems to work with PyQt5 and PyQt6.

image

firai:
Replace `lambda` in syntax highlighting style `QAction` with partial objects in `init_view_menu()` for compatibility with PySide2 and Pyside6 (without breaking compatibility with PyQt5 and PyQt6).

Fixes #537.

### DIFF
--- a/qtconsole/mainwindow.py
+++ b/qtconsole/mainwindow.py
@@ -6,6 +6,7 @@ common actions.
 
 # Copyright (c) Jupyter Development Team.
 # Distributed under the terms of the Modified BSD License.
+from functools import partial
 import sys
 import webbrowser
 from threading import Thread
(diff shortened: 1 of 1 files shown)


firai:
P.S. Not sure if this is something that could be shimmed in qtpy as well, in case other projects run into the same thing.

ccordoba12:
@firai, for next time, we don't require to maintain PRs branches always updated with master. So there's no need to keep merging them after merging other PRs.

PR #536 was a special case because tests were failing. That's basically our main requirement for a PR to be merged.

ccordoba12 (commented):
Thanks for your help @firai!

ccordoba12 (approved):

firai:
I only remerged because one of the tests randomly failed after I changed the order of imports, which didn't make any sense. I merged master more as a measure to retrigger the tests.

ccordoba12:
I see, then you can leave the task of re-running our tests to us. They are a bit flaky, so failures happen sometimes.

PR #540, opened 2022-05-05
Fix test_other_output for Qt 6.3

firai:
@ccordoba12 I found the reasons why `test_other_output()` is failing under Qt6:

- The existing expected `toHtml()` output seems to be written for Qt <6.3. HTML output from Qt 6.3 also adds a style definition for `hr` (even though no `hr` tags are actually used). I have added the Qt 6.3 expected HTML and modified environment detection conditions so that the test can be run under versions both above and below 6.3. Let me know if the test should only expect to be run with Qt >=6.3 and the expected output for <6.3 should be dumped. Note that nothing about the actual output has been changed.
- `document` also sometimes gets garbage collected before the assertions are run, resulting in exceptions from Qt. I have pushed the variable assignments as late as possible to avoid this.

Let me know if you have any comments.

### DIFF
--- a/qtconsole/tests/test_jupyter_widget.py
+++ b/qtconsole/tests/test_jupyter_widget.py
@@ -74,11 +74,12 @@ class TestJupyterWidget(unittest.TestCase):
         ))
 
         # Check proper syntax highlighting
-        if QT6 and parse(QT_VERSION) < parse('6.3'):
+        if QT6 and parse(QT_VERSION) >= parse('6.3'):
             html = (
                 '<!DOCTYPE HTML PUBLIC "-//W3C//DTD HTML 4.0//EN" "http://www.w3.org/TR/REC-html40/strict.dtd">\n'
                 '<html><head><meta name="qrichtext" content="1" /><meta charset="utf-8" /><style type="text/css">\n'
                 'p, li { white-space: pre-wrap; }\n'
+                'hr { height: 1px; border-width: 0; }\n'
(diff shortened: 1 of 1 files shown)


firai:
I realized that my initial characterization that the additional `hr` style is a difference between PySide and PyQt was not correct. It's actually a difference between Qt >=6.3 and <6.3. I have updated my comment above and the pull request to reflect this. As mentioned above, let me know if only keeping the Qt >=6.3 expected output is preferred  (i.e. the Qt <6.3 expected output should be dumped).

qtconsole/tests/test_jupyter_widget.py
```diff
@@ -72,11 +74,25 @@ def test_other_output(self):
         ))
 
         # Check proper syntax highlighting
-        if QT6:
+        if QT6 and parse(QT_VERSION) < parse('6.3'):
```
ccordoba12, line 77: Please change this so that the Qt >= 6.3 case comes first and then the other one. I think that makes this test easier to read,

ccordoba12 (commented):
Thanks a lot for your help with this @firai! I left a small comment for you, otherwise looks good to me.

ccordoba12 (approved):
Looks good to me now, thanks @firai!

firai:
@ccordoba12 By the way, is there a convention of adding anything directly called into the requirements in this project, or are implied dependencies sufficient? `packaging` is a dependency of `qtpy` and `setuptools`, so it's unlikely we would run into issues. But we could also add `packaging` explicitly to `extras_require` in `setup.py`.

ccordoba12:
We expect transitive dependencies of our dependencies to be available, so we don't declare them.

Issue #216, opened 2017-05-27
reverse-i-search (ctrl-r)


anntzer:
It would be useful if the qtconsole has a shortcut for reverse-searching the history, similar to readline and ipython (via ptk)'s ctrl-r (reverse-i-search).

modulusmath:
+1 - That would make qtconsole even more awesome.

franciobr:
I really miss that. It would be great.

iskreng:
I so wish that this is implemented! It's so inefficient not having it!

Is someone actually doing something about it?

ccordoba12:
No, it's not been worked on, sorry.

griano:
This would be great to have, since it is expected behavior for anyone that works in Linux shells. However, you can get similar behavior if you start typing and then click "UP".

khughitt:
+1. I'm caught off-guard every time I try to use `cntrl-r` to go through the history. Out of all the terminals/REPL's I jump between, this is the only one that breaks that expectation.

pizdjuk:
I can not understand, how this important feauture still not realized.

wmvanvliet:
> I can not understand, how this important feauture still not realized.

You're very welcome to make a PR for it.

pizdjuk:
what is PR?

ccordoba12:
It's a pull request.

dalthviz:
Maybe PR #622 could be thought as something that could resolve this @ccordoba12 ?

boffi:
> It would be useful if the qtconsole has a shortcut for reverse-searching the history, similar to readline and ipython (via ptk)'s ctrl-r (reverse-i-search).

Oh yes, please pretty please…

rmccampbell:
Interestingly, the help text (in Show QtConsole Help) says this:

> - ``Shift-Up``: search backwards through the history (like ``Control-r`` in bash).
> - ``Shift-Down``: search forwards through the history.

But these don't actually work, instead shift+up just selects the above text. This does imply that it was implemented at some point though, or at least was planned to be.

PR #555, opened 2022-08-09
Please, implement a backward search (Ctrl+R) feature in a console

pizdjuk:
Copy of https://github.com/jupyter/qtconsole/issues/216

=============
It would be useful if the qtconsole has a shortcut for reverse-searching the history, similar to readline and ipython (via ptk)'s ctrl-r (reverse-i-search).
=============

### DIFF
--- a/.github/workflows/linux-tests.yml
+++ b/.github/workflows/linux-tests.yml
@@ -20,7 +20,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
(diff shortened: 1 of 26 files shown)


ccordoba12:
What we meant on issue #216 is that, if you want to see this feature in Qtconsole soon, you need to implement it yourself because we don't have time to do it right now.

What you're doing here is simply not that, so I'm going to close this PR.

PR #568, opened 2023-01-14
Remove usage of `QVersionNumber` because it doesn't work as expected in PySide6

### LINKED ISSUE #567: Crash on PySide6 due to version checking
When I try to run qtconsole on macOS with PySide6, I get this error:

```
  File "/Users/john/miniconda3/envs/idaes/bin/jupyter-qtconsole", line 8, in <module>
    sys.exit(main())
  File "/Users/john/miniconda3/envs/idaes/lib/python3.10/site-packages/qtconsole/qtconsoleapp.py", line 467, in main
    JupyterQtConsoleApp.launch_instance()
  File "/Users/john/miniconda3/envs/idaes/lib/python3.10/site-packages/jupyter_core/application.py", line 277, in launch_instance
    return super().launch_instance(argv=argv, **kwargs)
  File "/Users/john/miniconda3/envs/idaes/lib/python3.10/site-packages/traitlets/config/application.py", line 1040, in launch_instance
    app.initialize(argv)
  File "/Users/john/miniconda3/envs/idaes/lib/python3.10/site-packages/traitlets/config/application.py", line 113, in inner
    return method(app, *args, **kwargs)
  File "/Users/john/miniconda3/envs/idaes/lib/python3.10/site-packages/qtconsole/qtconsoleapp.py", line 419, in initialize
    v_5_15_2 = QtCore.QVersionNumber.fromString('5.15.2')[0]
TypeError: 'PySide6.QtCore.QVersionNumber' object is not subscriptable
```

The issue comes from these lines, I assume the `QVersionNumber` object is different than expected in PySide6.

```
        # Fixes launching issues with Big Sur
        # https://bugreports.qt.io/browse/QTBUG-87014, fixed in qt 5.15.2
        #if sys.platform == 'darwin':
        #    v_5_15_2 = QtCore.QVersionNumber.fromString('5.15.2')[0]
        #    v_current = QtCore.QVersionNumber.fromString(QT_VERSION)[0]
        #    if v_current < v_5_15_2:
        #        os.environ['QT_MAC_WANTS_LAYER'] = '1'
```

psobolewskiPhD:
This PR closes #567 
If using pyside6 backend, `QVersionNumber.fromString('5.15.2')` returns a `PySide6.QtCore.QVersionNumber` and as a result it is not subscriptable, so launching the console results in a `TypeError`.
Because the code refers to Qt5 versions, I've added a check that the backend isn't Qt6 prior to the version checking.
On my mac this console now works in pyside6 6.4.2, though there is a warning:
```
python3.10/site-packages/jupyter_client/threaded.py:73: RuntimeWarning: ZMQStream only supports the base zmq.Socket class.

                Use zmq.Socket(shadow=other_socket)
                or `ctx.socket(zmq.SUB, socket_class=zmq.Socket)`
                to create a base zmq.Socket object,
                no matter what other kind of socket your Context creates.
                
  self.stream = zmqstream.ZMQStream(self.socket, self.ioloop)
```
I think that is unrelated to this PR and probably requires another fix.

### DIFF
--- a/qtconsole/qtconsoleapp.py
+++ b/qtconsole/qtconsoleapp.py
@@ -10,9 +10,9 @@ input, there is no real readline support, among other limitations.
 import os
 import signal
 import sys
-from packaging.version import parse
 from warnings import warn
 
+from packaging.version import parse
 # If run on Windows:
 #
 # 1. Install an exception hook which pops up a message box.
--- a/qtconsole/qtconsoleapp.py
+++ b/qtconsole/qtconsoleapp.py
@@ -10,7 +10,7 @@ input, there is no real readline support, among other limitations.
 import os
 import signal
 import sys
-from packaging import parse
+from packaging.version import parse
 from warnings import warn
 
 # If run on Windows:
--- a/qtconsole/qtconsoleapp.py
+++ b/qtconsole/qtconsoleapp.py
@@ -13,6 +13,7 @@ import sys
 from warnings import warn
 
 from packaging.version import parse
+
 # If run on Windows:
 #
 # 1. Install an exception hook which pops up a message box.
--- a/qtconsole/qtconsoleapp.py
+++ b/qtconsole/qtconsoleapp.py
@@ -57,7 +57,7 @@ if os.name == 'nt':
     except AttributeError:
         pass
 
-from qtpy import QtCore, QtGui, QtWidgets, QT_VERSION
+from qtpy import QtCore, QtGui, QtWidgets, QT_VERSION, QT6
 
 from traitlets.config.application import boolean_flag
 from traitlets.config.application import catch_config_error
@@ -415,7 +415,7 @@ class JupyterQtConsoleApp(JupyterApp, JupyterConsoleApp):
     def initialize(self, argv=None):
         # Fixes launching issues with Big Sur
         # https://bugreports.qt.io/browse/QTBUG-87014, fixed in qt 5.15.2
-        if sys.platform == 'darwin':
+        if sys.platform == 'darwin' and not QT6:
             v_5_15_2 = QtCore.QVersionNumber.fromString('5.15.2')[0]
             v_current = QtCore.QVersionNumber.fromString(QT_VERSION)[0]
             if v_current < v_5_15_2:
--- a/qtconsole/qtconsoleapp.py
+++ b/qtconsole/qtconsoleapp.py
@@ -10,6 +10,7 @@ input, there is no real readline support, among other limitations.
 import os
 import signal
 import sys
+from packaging import parse
 from warnings import warn
 
 # If run on Windows:
@@ -57,7 +58,7 @@ if os.name == 'nt':
     except AttributeError:
         pass
 
-from qtpy import QtCore, QtGui, QtWidgets, QT_VERSION, QT6
+from qtpy import QtCore, QtGui, QtWidgets, QT_VERSION
 
 from traitlets.config.application import boolean_flag
 from traitlets.config.application import catch_config_error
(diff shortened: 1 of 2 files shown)


psobolewskiPhD:
Hmm, this PR shouldn't affect anything in the linux tests.
So I think the failed linux tests are likely because the ubuntu runners are now 22.04
https://github.com/actions/runner-images/issues/5490
I guess something needs to be changed in the system apt install part of the workflow?

akirchhoff-modular (approved):
As one data point, I was encountering this issue as well, and this PR fixes it.  The change looks straightforward and makes sense.  I hope this can be merged soon.

qtconsole/qtconsoleapp.py
```diff
@@ -415,7 +415,7 @@ def _init_asyncio_patch(self):
     def initialize(self, argv=None):
         # Fixes launching issues with Big Sur
         # https://bugreports.qt.io/browse/QTBUG-87014, fixed in qt 5.15.2
-        if sys.platform == 'darwin':
+        if sys.platform == 'darwin' and not QT6:
```
ccordoba12, line 418: I don't think this is the right solution @psobolewskiPhD because we'd probably have similar checks to thes one below in the future.

Therefore, I propose to use instead the `parse` function from the `packaging` module instead, as it was used in our tests until PR #570:

https://github.com/jupyter/qtconsole/blob/a2aa5a11c883f5fd648eb56c19bf463d1cad99b1/qtconsole/tests/test_jupyter_widget.py#L77

Let us know if you feel comfortable doing this change or you prefer for me to do it.

ccordoba12 (commented):
Thanks @psobolewskiPhD for your contribution!

qtconsole/qtconsoleapp.py
psobolewskiPhD: Hi @ccordoba12 do you mean in the lines below, replace `QtCore.QVersionNumber.fromString` with `parse`? Rather than just skipping that QT5 code on QT6 that is.

ccordoba12: Yep, that's exactly what I meant.

psobolewskiPhD: Looking at my comment in the issue I had thought about doing that, but then did the lazy way 🤣 
Anyhow, I think I've got it sorted—works for me locally with pyside 6.4.2

qtconsole/qtconsoleapp.py
```diff
@@ -10,6 +10,7 @@
 import os
 import signal
 import sys
+from packaging.version import parse
 from warnings import warn
```
ccordoba12, line 15:
```suggestion
from warnings import warn

from packaging.version import parse
```

This is a third-party package, so it needs to go in a separate section.

ccordoba12 (commented):
Besides my small suggestion below, please add an entry for `packaging` in `setup.py` because this is a new dependency. Specifically, you need to add it on this list:

https://github.com/jupyter/qtconsole/blob/55a8fb0ed3c4f8c274adef7ebb077b4157d58550/setup.py#L68-L77

And you also need to merge with master/rebase to get the fixes to our tests I did in PR #570

qtconsole/qtconsoleapp.py
psobolewskiPhD: Sorry I swore it was a standard library, my bad!

ccordoba12: No problem. You would expect it to be, but (sadly) it isn't.

ccordoba12 (approved):
Looks good to me now, thanks @psobolewskiPhD!

PR #570, opened 2023-03-06
Cast images width and height to int when trying to insert them

ccordoba12:
- This fixes an error in Python 3.10+ when those values come as floats from Seaborn or Matplotlib.
- See https://github.com/spyder-ide/spyder/issues/20597 for more context.

### DIFF
--- a/.github/workflows/linux-tests.yml
+++ b/.github/workflows/linux-tests.yml
@@ -9,6 +9,10 @@ on:
     - master
   workflow_dispatch:
 
+concurrency:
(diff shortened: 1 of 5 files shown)


ccordoba12:
@blink1073, could you restore regular merges for this repo? I don't like squash-merges because they make harder to detect regressions with `git blame`.

blink1073:
Done, and I made you an admin for the  repo.

ccordoba12:
Thanks @blink1073!

PR #574, opened 2023-03-29
Check if the iopub channel is not closed before flushing it

ccordoba12:
This is necessary to avoid the error reported in https://github.com/spyder-ide/spyder/issues/20381#issuecomment-1482990746.

### DIFF
--- a/qtconsole/client.py
+++ b/qtconsole/client.py
@@ -45,6 +45,10 @@ class QtZMQSocketChannel(ThreadedZMQSocketChannel, SuperQObject):
         # Emit the generic signal.
         self.message_received.emit(msg)
 
+    def closed(self):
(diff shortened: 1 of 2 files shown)


qtconsole/client.py
```diff
@@ -45,6 +45,10 @@ def call_handlers(self, msg):
         # Emit the generic signal.
         self.message_received.emit(msg)
 
+    def closed(self):
+        """Check if the channel is closed."""
+        return self.stream.closed()
```
minrk, line 50: After it finishes closing, the channel sets it to None, so handle that, too:

```suggestion
        return self.stream is None or self.stream.closed()
```

minrk (approved):
Yup, that's exactly what I was talking about.

Issue #378, opened 2019-09-18
"MultiKey" treatment


boffi:
I'm using QT Console, as shipped from Anaconda, with X on Linux.
> Linux debian 5.2.0-2-amd64 #1 SMP Debian 5.2.9-2 (2019-08-21) x86_64 GNU/Linux (sid)
> conda 4.7.11
> Jupyter QtConsole 4.5.4
> Python 3.7.4 (default, Aug 13 2019, 20:35:49) 
> Type 'copyright', 'credits' or 'license' for more information
> IPython 7.7.0 -- An enhanced Interactive Python. Type '?' for help.

When I press "MultiKey" to start a multi key XCompose sequence the QT Console inserts a Unicode character 'EN DASH' (U+2013).
The expected behaviour in a X program is to wait for following keystrokes and possibly insert a character when a valid sequence is detected or nothing when a valid sequence is no more possible.
I notice that the same behaviour is shown also from the console as installed via a Debian Package (`python3-qtconsole 4.3.1-1`)
> Jupyter QtConsole 4.3.1
> Python 3.7.4+ (default, Sep  4 2019, 08:03:05) 
> IPython 5.8.0 -- An enhanced Interactive Python.

ccordoba12:
You're welcome to submit a pull request to add support for this because I think we don't have it (I don't even understand what you mean by it).

PR #575, opened 2023-04-04
Add missing `closed` method to `QtInProcessChannel`

rayosborn:
In PR #574, a new function, `closed()` was added to the `QtZMQSocketChannel` class in order to check whether a channel was closed before flushing the stream. This must also be added to the `QtInProcessChannel` to ensure compatibility. Since in-process channels are never closed, this function should always return the value `False`.

This PR corrects the NeXpy issue [#378](https://github.com/nexpy/nexpy/issues/398) using the latest development version of qtconsole.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -21,3 +21,4 @@ __pycache__
 .#*
 .coverage
 .pytest_cache
+.vscode
--- a/qtconsole/inprocess.py
+++ b/qtconsole/inprocess.py
@@ -56,6 +56,10 @@ class QtInProcessChannel(SuperQObject, InProcessChannel):
         super().flush()
         self.process_events()
 
+    def closed(self):
+        """ Function to ensure compatibility with the QtZMQSocketChannel."""
+        return False
+
 
 class QtInProcessHBChannel(SuperQObject, InProcessHBChannel):
     # This signal will never be fired, but it needs to exist
--- a/qtconsole/tests/test_inprocess_kernel.py
+++ b/qtconsole/tests/test_inprocess_kernel.py
@@ -1,0 +1,31 @@
+"""Test QtInProcessKernel"""
+
+# Copyright (c) Jupyter Development Team.
+# Distributed under the terms of the Modified BSD License.
+
+import unittest
+
+from qtconsole.inprocess import QtInProcessKernelManager
+
+
+class InProcessTests(unittest.TestCase):
+
+    def setUp(self):
+        """Open an in-process kernel."""
+        self.kernel_manager = QtInProcessKernelManager()
+        self.kernel_manager.start_kernel()
+        self.kernel_client = self.kernel_manager.client()
+
+    def tearDown(self):
+        """Shutdown the in-process kernel. """
+        self.kernel_client.stop_channels()
+        self.kernel_manager.shutdown_kernel()
+
+    def test_execute(self):
+        """Test execution of shell commands."""
+        # check that closed works as expected
+        assert not self.kernel_client.iopub_channel.closed()
+        
+        # check that running code works
+        self.kernel_client.execute('a=1')
+        assert self.kernel_manager.kernel.shell.user_ns.get('a') == 1


ccordoba12:
Thanks for your contribution @rayosborn! Perhaps this is too much to ask, but is there any chance you could add a test for the InProcess integration? That way we will be sure of not breaking it in the future.

rayosborn:
Thanks for the suggestion. I will have a look at the current test functions to see if they could be adapted for use with the in-process kernel. I can't guarantee I have the right expertise, but I agree that it could be a useful addition to the test suite.

adam-urbanczyk:
@ccordoba12 CQ-editor is also suffering from this issue. AFAICT one should parameterize some tests with different KernelManagers (i.e. include `InProcessKernelManager`), e.g. here:

https://github.com/jupyter/qtconsole/blob/949c0ec2020f80971e082121d4571fb11c8d928f/qtconsole/tests/test_comms.py#L13

What would be the simplest way to achieve this with the current code base? I only know how to do this nicely with pytest.

ccordoba12:
A simple test that checks that the console can run code with the in-process kernel manager would do for now. No need to try to make our entire test suite run with it.

adam-urbanczyk:
@rayosborn AFAICT something like this should be enough. Could you try?

```python
class InProcessTests(unittest.TestCase):

    def setUp(self):
        """Open a kernel."""
        self.kernel_manager = QtInProcessKernelManager()
        self.kernel_manager.start_kernel()
        self.kernel_client = self.kernel_manager.client()
        self.kernel_client.start_channels(shell=True, iopub=True)
        self.blocking_client = self.kernel_client.blocking_client()
        self.blocking_client.start_channels(shell=True, iopub=True)
        self.comm_manager = self.kernel_client.comm_manager

        # Check if client is working
        self.blocking_client.execute('print(0)')
        try:
            self._get_next_msg()
            self._get_next_msg()
        except TimeoutError:
            # Maybe it works now?
            self.blocking_client.execute('print(0)')
            self._get_next_msg()
            self._get_next_msg()


    def tearDown(self):
        """Close the kernel."""
        if self.kernel_manager:
            self.kernel_manager.shutdown_kernel(now=True)
        if self.kernel_client:
            self.kernel_client.shutdown()
            
    def _get_next_msg(self, timeout=10):
        # Get status messages
        timeout_time = time.time() + timeout
        msg_type = 'status'
        while msg_type == 'status':
            if timeout_time < time.time():
                raise TimeoutError
            try:
                msg = self.blocking_client.get_iopub_msg(timeout=3)
                msg_type = msg['header']['msg_type']
            except Empty:
                pass
        return msg

    def test_execute(self):
        self.kernel_client.execute('a=1')
        assert self.kernel_manager.kernel.shell.user_ns.get('a') == 1

```

rayosborn:
@adam-urbanczyk, I get errors unless I remove the initialization of `blocking_client` and the calls to shutdown the kernel and client. Unfortunately, it is then not sensitive to whether the new `closed` function is present or not. I'm not so familiar with using the `unittest` framework so there may be things I need to initialize first. Does it work for you?

adam-urbanczyk:
This seems to be better:
```python
    def setUp(self):
        """Open a kernel."""
        self.kernel_manager = QtInProcessKernelManager()
        self.kernel_manager.start_kernel()
        self.kernel_client = self.kernel_manager.client()

    def tearDown(self):
        pass

    def test_execute(self):
        
        # check that closed works as expected
        assert not self.kernel_client.iopub_channel.closed()
        
        # check that running code works
        self.kernel_client.execute('a=1')
        assert self.kernel_manager.kernel.shell.user_ns.get('a') == 1
```

Maybe commit just commit some test and ask the owners for feedback? I'm not familiar with the codebase enough to be able to tell what is sufficient.

rayosborn:
The trimmed down version seems to work. I got the test to fail by removing the `closed` function from `in process.py`, so it at least catches the reason for this PR. @adam-urbanczyk, thanks for doing the work. 

I was able to add the following to the `tearDown` function. 
```
def tearDown(self):
    self.kernel_client.stop_channels()
    self.kernel_manager.shutdown_kernel()
```
I am not sure why that failed before but I have these lines in the NeXpy shutdown code. I presumably copied them from another package, but it was a long time ago and I'm not sure why they are needed. They emit some `stopped` signals and 'join' the `ioPub` thread. It's possible that other applications need this, so I will leave them in for now.

I will add this to the test suite in the current PR, assuming you have no objection.

ccordoba12 (approved):
Looks good to me now. Thanks @rayosborn and @adam-urbanczyk for your help with this!

ccordoba12:
I just released 5.4.3 with this change.

Issue #576, opened 2023-04-06
AttributeError: 'QtInProcessChannel' object has no attribute 'closed'


twizmwazin:
After commit ff57647679e1b20be56acaf90273e36f74d3601d, FrontendWidget calls `closed` on `self.kernel_client.iopub_channel`. This commit introduced that `closed` attribute to QtZMQSockerChannel. No such attribute exists for `QtInProcessChannel` however, so this exception is raised.

ccordoba12:
Correct. Since this issue is no longer valid, I'm going to close it.

PR #577, opened 2023-04-06
Implement .closed() for QtInProcessChannel

twizmwazin:
Possible fix for #576, this implements a .closed() for QtInProcessChannel. Another possible fix that might be more universal would be to instead check if iopub_channel has a closed attribute.


twizmwazin:
Just realized this is a duplicate of  #575.

PR #589, opened 2023-10-30
Remove Support for Python 2 Lexing

### LINKED ISSUE #588: Stop using `IPythonLexer` from ` IPython.lib.lexers`
In release 8.17.0 the ipython remove `IPythonLexer` from `IPython.lib.lexers` as cleanup of python 2 code. However qtconsole still import its: 

https://github.com/jupyter/qtconsole/blob/2accd52e7af226bc857a89099260e55c5740efe1/qtconsole/jupyter_widget.py#L17

That leads to crash on import time. 

I have opened issue in python repository:
https://github.com/ipython/ipython/issues/14227

But long term solution should be to not depend on `IPythonLexer`, especially that project ins python 3.7+.

Carreau:
I don't think there is much reasons to lex Python 2 with such a recent QtConsole/IPython.

closes #588

### DIFF
--- a/qtconsole/jupyter_widget.py
+++ b/qtconsole/jupyter_widget.py
@@ -14,7 +14,7 @@ from warnings import warn
 
 from qtpy import QtCore, QtGui
 
-from IPython.lib.lexers import IPythonLexer, IPython3Lexer
+from IPython.lib.lexers import IPython3Lexer
 from pygments.lexers import get_lexer_by_name
 from pygments.util import ClassNotFound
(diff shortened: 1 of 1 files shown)


ccordoba12:
@Carreau, thanks for your help with this! The thing is we still support Python 2 kernels in Spyder and some people rely on them. So, we need to keep support for it until our next major version, to be released at the beginning of next year.

ccordoba12 (approved):
Thanks @Carreau!

PR #590, opened 2023-11-10
Test usage of proactor event loop

blink1073:
Research for [discuss.python.org/t/removing-the-asyncio-policy-system-asyncio-set-event-loop-policy-in-python-3-15/37553/26](https://discuss.python.org/t/removing-the-asyncio-policy-system-asyncio-set-event-loop-policy-in-python-3-15/37553/26)

### DIFF
--- a/qtconsole/qtconsoleapp.py
+++ b/qtconsole/qtconsoleapp.py
@@ -397,21 +397,21 @@ class JupyterQtConsoleApp(JupyterApp, JupyterConsoleApp):
         FIXME: if/when tornado supports the defaults in asyncio,
                remove and bump tornado requirement for py38
         """
-        if sys.platform.startswith("win") and sys.version_info >= (3, 8):
-            import asyncio
-            try:
-                from asyncio import (
-                    WindowsProactorEventLoopPolicy,
-                    WindowsSelectorEventLoopPolicy,
-                )
-            except ImportError:
-                pass
-                # not affected
-            else:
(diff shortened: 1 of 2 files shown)


blink1073:
@ccordoba12 do you have any concerns about this?

blink1073:
For context, the ability to set an event loop policy might be going away.

ccordoba12:
I thought this was necessary on Windows for Python 3.8+. How did it get fix?

blink1073:
Hey @ccordoba12, tornado 6.1 added support for the proactor [loop](https://github.com/tornadoweb/tornado/blob/a6dfd70d7a6398b2187021df63ca77daa5781e5e/docs/releases/v6.1.0.rst#general-changes), by starting a thread. The `jupyter_server` test suite runs in about 10 min with the selector loop, and 15 min with the proactor loop.

Guido suggested that if we can switch from `ioloop.start()` to calling `asyncio.run(loop_factory=...)` ourselves, we could retain the proactor loop.

ccordoba12:
@blink1073, thanks for the explanation. I must confess my ignorance about most asyncio related stuff because I haven't had time to learn about it.

So, if you consider this is no longer necessary then I'm ok with removing it.

blink1073:
> So, if you consider this is no longer necessary then I'm ok with removing it.

I still need to do more testing to make sure we're future-proof (if they take away `set_event_loop_policy`, but thanks!

PR #592, opened 2023-11-14
Fix some utility functions

### LINKED ISSUE #591: TypeError due to wrong spelling in keyword
The call to `compute_item_matrix` function in https://github.com/jupyter/qtconsole/blob/master/qtconsole/completion_html.py#L315 uses the `displaywidth` keyword, but the implementation in https://github.com/jupyter/qtconsole/blob/master/qtconsole/util.py#L144 has is spelled as `displaywith` (no second d). This cause a type error when called in the code

fdobad:
Also required by columnize, line 223, utils.py

Fixes #591.

### DIFF
--- a/qtconsole/util.py
+++ b/qtconsole/util.py
@@ -204,7 +204,7 @@ def compute_item_matrix(items, empty=None, *, separator_size=2, displaywidth=80)
     )
 
 
-def columnize(items):
+def columnize(items, separator="  ", displaywidth=80):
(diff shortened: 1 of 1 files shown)


wmvanvliet (approved):

wmvanvliet:
closes #591

ccordoba12 (approved):
Thanks @fdobad!

ccordoba12:
I just released 5.5.1 with this fix.

PR #596, opened 2023-11-25
Update rtd yaml file and fix rtd build

willingc:


ccordoba12:
Thanks @willingc for your help with this!

PR #601, opened 2024-02-21
Use string representation for QKeySequence construction

mborgerson:
This fixes an issue in PySide6 v6.6.2 where key sequence construction with the | operator is broken, and improves readability. 
Ideally this would be fixed in PySide6 or qtpy to retain original behavior, but since it also improves readability I think it makes sense to fix it here too.

Broken with PySide6 v6.6.2:

```py
from qtpy import QtCore
QtCore.Qt.AltModifier | QtCore.Qt.Key.Key_C
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
Cell In[2], line 1
----> 1 QtCore.Qt.AltModifier | QtCore.Qt.Key.Key_C

TypeError: unsupported operand type(s) for |: 'KeyboardModifier' and 'Key'
```

### DIFF
--- a/qtconsole/frontend_widget.py
+++ b/qtconsole/frontend_widget.py
@@ -189,9 +189,8 @@ class FrontendWidget(HistoryConsoleWidget, BaseFrontendMixin):
 
         # Configure actions.
         action = self._copy_raw_action
-        key = QtCore.Qt.CTRL | QtCore.Qt.SHIFT | QtCore.Qt.Key_C
         action.setEnabled(False)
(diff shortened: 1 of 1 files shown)


mborgerson:
Hi @ccordoba12, would you mind taking a look at this patch? The failing CI appears to be unrelated to the patch.

ccordoba12 (approved):
Looks good to me, thanks @mborgerson!

(Sorry for the delay in reviewing your work. This one just missed my radar).

PR #602, opened 2024-05-03
Use new format for instersphinx_mapping

Carreau:
The current format is deprecated since sphinx 6.2 and will be removed in the future.


ccordoba12 (approved):
Thanks @Carreau!

Issue #553, opened 2022-07-06
Carriage Return Inconsistency


poipoi300:
# Context
I was writing a simple progress indicator for use internally in Spyder and found out that carriage returns don't always work, meaning that trying to print over lines in the console sometimes appends the new text instead. When I discovered this, I took out Spyder from the equation and tested directly in Qt Console. I experienced the same problem. I then updated Qt Console to the latest version (5.3.1), and the problem got better, however it still occurred and a new issue popped up where text would be updated much less often (sometimes not at all until the final output). I have tested this in the windows command line, and it behaves exactly as expected. I have tested this on 2 systems, both running windows (10 and 11).

### Expected behavior
Carriage returns always return the cursor to the start of the line and the console's output is updated for each new print.
### Actual behavior
Carriage returns are sometimes not taken into account, and text is appended when it should not be. The console's output updates are also delayed.
### Reproducing the behavior
```py
# If this snippet does not show issues, run it more times or add iterations.

for i in range(1000000):
    print(f'{i} ', end = '\r') # Flushing makes the issue worse 

# Example output:
# 999999 142189 773780
```

#### IPython.sys_info() output
```js
{'commit_hash': 'e321e760a',
 'commit_source': 'installation',
 'default_encoding': 'cp1252',
 'ipython_path': 'C:\\ProgramData\\Anaconda3\\Lib\\site-packages\\IPython',
 'ipython_version': '7.31.1',
 'os_name': 'nt',
 'platform': 'Windows-10-10.0.22000-SP0',
 'sys_executable': 'C:\\ProgramData\\Anaconda3\\python.exe',
 'sys_platform': 'win32',
 'sys_version': '3.9.12 (main, Apr  4 2022, 05:22:27) [MSC v.1916 64 bit '
                '(AMD64)]'}
```

ccordoba12:
Hey @poipoi300, thanks for reporting. This is basically a duplicate of issue #350.

It'd be great if you could give us a hand to improve our ANSI pre-processor. On that issue you can find the minimal info to do it, in case you are interested.

Issue #272, opened 2018-02-16
Strange print using \r 


sylvain-bougnoux:
Any of these do not work (with different outputs), w or w/o flush=True

for i in range(10000): print(i,end='\r')
for i in range(10000): print('\r',i,end='')
for i in range(10000): print('\r',i,end=' ')

last version. Python 3.6.3-64, windows

sylvain-bougnoux:
Sure sorry about that.
```
for i in range(10000): print(i,end='\r')
9999202945649130
```
```
for i in range(10000): print('\r',i,end='')
 9999 1109 358371729503
```
```
for i in range(10000): print('\r',i,end=' ')
 9999  1175  2400 3590   5981 7328 8477 
```
Seems like the same issue (a synchronization issue of flush) poping differently depending on timing.
I see it through spyder (which uses a qtconsole).

takluyver:
Thanks. cc @minrk - is there any chance that could be an issue with the io-capturing machinery in the kernel? I remember it involves a fair bit of complexity now.

Also, @sylvain-bougnoux , what version of ipykernel do you have?

sylvain-bougnoux:
last ones
ipykernel is 4.8.2
qtconsole is 4.3.1
ipython is 6.2.1

PR #607, opened 2024-05-27
Fix Carriage Return Handling in QtConsole

TheMatt2:
Fix the handling of carriage return so it is not ignored at the end of a print line.

This allows for progress bars that rely on adding a carriage return at the end of the line such as:

```python
print("Progress 55%", end = "\r")
```

tl;dr
Preserve position of cursor so ANSI commands such as \r that modify position will preserve the position they specified over multiple calls. This is particularly relevant if time passes between print calls (~100 ms).

## Explanation

First of all, when you `print()` or call `sys.stdout.write()` in QtConsole, this ultimately causes the text to be passed to the frontend widget.

https://github.com/jupyter/qtconsole/blob/56e5a5ec5cfccf5afd098fe8bb5b2c558a1dc8f9/qtconsole/frontend_widget.py#L729

This will lead to the text by passed to the console widget, which will call the ANSI Processor in:

https://github.com/jupyter/qtconsole/blob/56e5a5ec5cfccf5afd098fe8bb5b2c558a1dc8f9/qtconsole/console_widget.py#L2085

The ANSI preprocessor already handles the carriage return, so code that that uses the carriage return in the middle of a print already worked.

```
print("Hello\rWorld")
World
```

This will even work for calls from multiple prints, but this *only* works because the text is internally buffered, and recombined so `print("Hello", end = "\r"); print("World")` is all combined into the text `"Hello\rWorld" by the time `_insert_plain_text()` is called.

However, `_insert_plain_text()` and `_flush_pending_stream()` both force the cursor position to the end of the line at the end of the call. This means that a `'\r'` at the end of a line is effectively ignored, _if_ something stops the text from being combined back together, such as more than ~100 ms passing between prints.

Time

This PR fixes this issue by updating the cursor position to the place it was put to by `_insert_plain_text()`.

Before this gets merged, I do want to make sure there are no odd situations where the cursor now not being the end of text leads to issues.

### Before

Before

### After

After

## Notes

Related Github Issues

 - #272
 - #553 
 - https://github.com/spyder-ide/spyder/issues/195
 - https://github.com/spyder-ide/spyder/issues/20388

This does not exactly address #350, not adding support for any new ANSI codes, just fixing the existing support for `'\r'`

### DIFF
--- a/qtconsole/console_widget.py
+++ b/qtconsole/console_widget.py
@@ -2519,6 +2519,9 @@ class ConsoleWidget(MetaQObjectHasTraits('NewBase', (LoggingConfigurable, superQ
 
         cursor = self._get_end_cursor()
 
+        # Memorize end to check if we actually add any prompt characters
+        prior_end_pos = cursor.position()
+
         # Save the current position to support _append*(before_prompt=True).
         # We can't leave the cursor at the end of the document though, because
         # that would cause any further additions to move the cursor. Therefore,
@@ -2557,7 +2560,16 @@ class ConsoleWidget(MetaQObjectHasTraits('NewBase', (LoggingConfigurable, superQ
                 self._prompt_html = None
 
         self._flush_pending_stream()
-        self._prompt_cursor.setPosition(self._get_end_pos() - 1)
+
+        current_end_pos = self._get_end_pos()
+        if prior_end_pos != current_end_pos:
+            # Set the prompt cursor to end minus 1, so long as
+            # the prompt was not blank
+            self._prompt_cursor.setPosition(current_end_pos - 1)
+        else:
+            # The prompt didn't move end, i.e. the prompt inserted exactly 0 characters
+            # Move cursor to end
+            self._prompt_cursor.setPosition(current_end_pos)
 
         if move_forward:
             self._append_before_prompt_cursor.setPosition(
--- a/qtconsole/console_widget.py
+++ b/qtconsole/console_widget.py
@@ -2519,9 +2519,6 @@ class ConsoleWidget(MetaQObjectHasTraits('NewBase', (LoggingConfigurable, superQ
 
         cursor = self._get_end_cursor()
 
-        # Memorize end to check if we actually add any prompt characters
-        prior_end_pos = cursor.position()
-
         # Save the current position to support _append*(before_prompt=True).
         # We can't leave the cursor at the end of the document though, because
         # that would cause any further additions to move the cursor. Therefore,
@@ -2560,16 +2557,7 @@ class ConsoleWidget(MetaQObjectHasTraits('NewBase', (LoggingConfigurable, superQ
                 self._prompt_html = None
 
         self._flush_pending_stream()
-
-        current_end_pos = self._get_end_pos()
-        if prior_end_pos != current_end_pos:
-            # Set the prompt cursor to end minus 1, so long as
-            # the prompt was not blank
-            self._prompt_cursor.setPosition(current_end_pos - 1)
-        else:
-            # The prompt didn't move end, i.e. the prompt inserted exactly 0 characters
-            # Move cursor to end
-            self._prompt_cursor.setPosition(current_end_pos)
+        self._prompt_cursor.setPosition(self._get_end_pos() - 1)
 
         if move_forward:
             self._append_before_prompt_cursor.setPosition(
--- a/qtconsole/console_widget.py
+++ b/qtconsole/console_widget.py
@@ -1028,7 +1028,7 @@ class ConsoleWidget(MetaQObjectHasTraits('NewBase', (LoggingConfigurable, superQ
             # Insert at current printing point
             # If cursor is before prompt jump to end, but only if there
             # is a prompt (before_prompt_pos != end)
-            if cursor.position() < self._append_before_prompt_pos \
+            if cursor.position() <= self._append_before_prompt_pos \
                     and self._append_before_prompt_pos != self._get_end_pos():
                 cursor.movePosition(QtGui.QTextCursor.End)
 
@@ -2562,6 +2562,9 @@ class ConsoleWidget(MetaQObjectHasTraits('NewBase', (LoggingConfigurable, superQ
         if move_forward:
             self._append_before_prompt_cursor.setPosition(
                 self._append_before_prompt_cursor.position() + 1)
+        else:
+            # cursor position was 0, set before prompt cursor
+            self._append_before_prompt_cursor.setPosition(0)
         self._prompt_started()
 
     #------ Signal handlers ----------------------------------------------------
--- a/qtconsole/console_widget.py
+++ b/qtconsole/console_widget.py
@@ -297,6 +297,10 @@ class ConsoleWidget(MetaQObjectHasTraits('NewBase', (LoggingConfigurable, superQ
         self._reading_callback = None
         self._tab_width = 4
 
+        # Cursor position of where to insert text.
+        # Control characters allow this to move around on the current line.
+        self._insert_text_cursor = self._control.textCursor()
+
         # List of strings pending to be appended as plain text in the widget.
         # The text is not immediately inserted when available to not
         # choke the Qt event loop with paint events for the widget in
@@ -695,6 +699,9 @@ class ConsoleWidget(MetaQObjectHasTraits('NewBase', (LoggingConfigurable, superQ
             # effect when using a QTextEdit. I believe this is a Qt bug.
             self._control.moveCursor(QtGui.QTextCursor.End)
 
+            # Advance where text is inserted
+            self._insert_text_cursor.movePosition(QtGui.QTextCursor.End)
+
     def export_html(self):
         """ Shows a dialog to export HTML/XML in various formats.
         """
@@ -710,6 +717,9 @@ class ConsoleWidget(MetaQObjectHasTraits('NewBase', (LoggingConfigurable, superQ
 
         # There is no prompt now, so before_prompt_position is eof
         self._append_before_prompt_cursor.setPosition(
+            self._get_end_cursor().position())
+
+        self._insert_text_cursor.setPosition(
             self._get_end_cursor().position())
 
         # The maximum block count is only in effect during execution.
@@ -998,7 +1008,7 @@ class ConsoleWidget(MetaQObjectHasTraits('NewBase', (LoggingConfigurable, superQ
         current prompt, if there is one.
         """
         # Determine where to insert the content.
-        cursor = self._control.textCursor()
+        cursor = self._insert_text_cursor
         if before_prompt and (self._reading or not self._executing):
             self._flush_pending_stream()
             cursor._insert_mode=True
@@ -1009,6 +1019,11 @@ class ConsoleWidget(MetaQObjectHasTraits('NewBase', (LoggingConfigurable, superQ
 
         # Perform the insertion.
         result = insert(cursor, input, *args, **kwargs)
+
+        # Remove insert mode tag
+        if hasattr(cursor, "_insert_mode"):
+            del cursor._insert_mode
+
         return result
 
     def _append_block(self, block_format=None, before_prompt=False):
@@ -1670,7 +1685,7 @@ class ConsoleWidget(MetaQObjectHasTraits('NewBase', (LoggingConfigurable, superQ
             text = self._get_last_lines_from_list(text, buffer_size)
         text = ''.join(text)
         t = time.time()
-        self._insert_plain_text(self._control.textCursor(), text, flush=True)
+        self._insert_plain_text(self._insert_text_cursor, text, flush=True)
         # Set the flush interval to equal the maximum time to update text.
         self._pending_text_flush_interval.setInterval(
             int(max(100, (time.time() - t) * 1000))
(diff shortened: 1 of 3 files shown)


TheMatt2:
Added regression test. Needed to run it manually, as the `test_scroll()` keeps failing (it worked once though)

ccordoba12:
Hey @TheMatt2, thanks for your contribution! Unfortunately, it doesn't seem the right approach if it breaks scrolling.

My suggestion is for you to rethink your approach so that `test_scroll` passes consistently.

TheMatt2:
> Hey @TheMatt2, thanks for your contribution! Unfortunately, it doesn't seem the right approach if it breaks scrolling.
> 
> My suggestion is for you to rethink your changes so that `test_scroll` passes consistently.

Let me rephrase. `test_scroll()` is breaking for me, _without_ my changes. I don't think my changes are impacting scroll, that test is failing on its own.

Working on it, to see if I can figure it out, and go through the CI results.

ccordoba12:
Ok, sorry for the misunderstanding. Since I haven't seen that test failing before, I assumed it was do to your changes.

TheMatt2:
So it seems the following error the CI reports is introduced by the PR, I'm looking into what the implications are.


```
FAILED qtconsole/tests/test_00_console_widget.py::test_scroll[True] - assert 1930 == 1170
```

The error I was referring to previously, is in my testing `test_00_console_widget.py::test_scroll()` hangs forever on MacOS, unrelated to the changes in this PR.

And on Windows, `test_00_console_widget.py::test_scroll()` is failing for me with an unrelated exception.

And it appears these failures were masking that there is an issue in this PR that needs to be addressed.
Suppose I may need to open issues for there others, but one thing at a time.

TheMatt2:
Updated approach to not break `test_scroll()`. While I am not sure exactly why this broke the test, the issue was printing was screwed up if the user clicked around during printing (evidently setTextCursor()` does not work how I thought it did.)

TheMatt2:
Change to draft, as `test_carriage_return` fails if `w._executing = False`. Want to resolve this before merging.

TheMatt2:
After much revision, found a way to deal with this that solves the issue.

The issue was text is added using `_insert_mode` in `cursor_widget.py`  if we are executing . The problem is insert mode breaks the \r handling, as we never "swallow" characters, only insert.

Solution is to only use `_insert mode` if text is being inserted on the line that holds the current prompt, as that is the only time we truly need this behavior.

This has been added to the test cases.


The way you could trigger this corner case where you still need insert mode is something like this

image

TheMatt2:
At this point, the issue is solved, and all test cases pass, but I have one last edge case I want to chase down before removing the draft tag.

TheMatt2:
Added fix for an edge case where switching between printing before the prompt, and printing after the prompt caused the `self._prompt_sep` to be printed in the wrong position.

All tests past... but there's a remaining issue if the prompt is a blank character... i.e. `input('')`. Still a draft.

TheMatt2:
So if you run the following code fragment, you can hang the GUI, and it prints the error `QTextCursor::setPosition: Position '429' out of range`

This occurs on the main branch right now, not actually caused by any change I've made here. Attempting a fix.

image

TheMatt2:
So I've confirmed this is a separate issue that doesn't require any changes to this PR.
I tried a fix for this, but it that fix actually broke  like it is actually going to be a little involved.

Before digging too much into that, I would actually like some feedback on the change in this PR, as this seems to now all being working (took a while though).

dalthviz:
Hi @TheMatt2 thank you for all you work here and sorry for the late response! Could it be possible for you to merge with latest main so this can get green CI (some tests fixes have been done over there)? Also, I gave a quick check to this PR manually and seems like things are working as expected :+1: Will try to do a more in depth code review now but if there are any specific changes that you think need more attention let us know!

ccordoba12:
> @ccordoba12 Thoughts?

Sorry for dropping the ball here. @dalthviz will take it from here, he's also a maintainer of this repo.

qtconsole/console_widget.py
```diff
@@ -998,18 +1008,40 @@ def _append_custom(self, insert, input, before_prompt=False, *args, **kwargs):
...
+                if cursor.selection().toPlainText() != '\n':
+                    cursor._insert_mode = True
+                cursor.movePosition(QtGui.QTextCursor.Left)
         else:
+            # Insert at current printing point
+            # If cursor is before prompt jump to end, but only if there
+            # is a prompt (before_prompt_pos != end)
+            if cursor.position() <= self._append_before_prompt_pos \
+                    and self._append_before_prompt_pos != self._get_end_pos():
+                cursor.movePosition(QtGui.QTextCursor.End)
```
dalthviz, line 1033: A format suggestion:
```suggestion
            if (
                cursor.position() <= self._append_before_prompt_pos
                and self._append_before_prompt_pos != self._get_end_pos()
            ):
                cursor.movePosition(QtGui.QTextCursor.End)
```

qtconsole/console_widget.py
```diff
@@ -998,18 +1008,40 @@ def _append_custom(self, insert, input, before_prompt=False, *args, **kwargs):
...
+
             if insert != self._insert_plain_text:
                 self._flush_pending_stream()
-            cursor.movePosition(QtGui.QTextCursor.End)
 
         # Perform the insertion.
         result = insert(cursor, input, *args, **kwargs)
+
+        # Remove insert mode tag
+        if hasattr(cursor, "_insert_mode"):
```
dalthviz, line 1042: A format suggestion:
```suggestion
        if hasattr(cursor, '_insert_mode'):
```

qtconsole/console_widget.py
```diff
@@ -1660,10 +1692,7 @@ def _on_flush_pending_stream_timer(self):
         """ Flush the pending stream output and change the
         prompt position appropriately.
         """
```
dalthviz, line 1694: Probably we should update this method docstring to just be something like:

```python
""" Flush the pending stream output."""
```

qtconsole/frontend_widget.py
```diff
@@ -728,7 +728,7 @@ def restart_kernel(self, message, now=False):
 
     def append_stream(self, text):
         """Appends text to the output stream."""
-        self._append_plain_text(text, before_prompt=True)
+        self._append_plain_text(text, before_prompt = True)
```
dalthviz, line 731: I think this could be reverted since the equal is for a `kwarg`. So something like:
```suggestion
        self._append_plain_text(text, before_prompt=True)
```

qtconsole/tests/test_00_console_widget.py
```diff
@@ -448,7 +486,7 @@ def test_prompt_cursors(self):
                          w._prompt_pos - len(w._prompt))
 
         # insert some text before the prompt
-        w._append_plain_text('line', before_prompt=True)
+        w._append_plain_text('line', before_prompt = True)
```
dalthviz, line 489: Same as above, I think we could revert this change since the equal is for a `kwarg`.  So something like:
```suggestion
        w._append_plain_text('line', before_prompt=True)
```

qtconsole/tests/test_00_console_widget.py
```diff
@@ -371,6 +371,44 @@ def test_erase_in_line(self):
...
+        self.assert_text_equal(w._get_cursor(), '\u2029prompt>')
+
+        test_inputs = ['Hello\n', 'World\r',
+                       '*' * 10, '\r',
+                       '0', '1', '2', '3', '4',
+                       '5', '6', '7', '8', '9',
+                       '\r\n']
+
+        for text in test_inputs:
+            w._append_plain_text(text, before_prompt = True)
```
dalthviz, line 394: A format suggestion:
```suggestion
            w._append_plain_text(text, before_prompt=True)
```

qtconsole/tests/test_00_console_widget.py
```diff
@@ -371,6 +371,44 @@ def test_erase_in_line(self):
...
+        self.assert_text_equal(w._get_cursor(),
+            "Hello\u20290123456789\u2029\u2029prompt>")
+
+        # Print after prompt
+        w._executing = True
+        test_inputs = ['\nF', 'o', 'o',
+                       '\r', 'Bar', '\n']
+
+        for text in test_inputs:
+            w._append_plain_text(text, before_prompt = False)
```
dalthviz, line 406: A format suggestion:
```suggestion
            w._append_plain_text(text, before_prompt=False)
```

dalthviz (commented):
Thank you again @TheMatt2 for your work here! Left a comment to update a method docstring and several suggestions for format nitpicks. Other than that this LGTM :+1:

qtconsole/console_widget.py
TheMatt2: Now that you point that out, I see the rest of the code is rather consistent on this, changing.

I'm curious, is there specific guidelines for using single/double quotes?

qtconsole/console_widget.py
TheMatt2: I do appreciate the code review. And I follow avoiding the \ when possible as a general guideline.
Though, it seems to be that the majority of multi-line if statements in this file use backslashes as opposed to explicitly encapsulated with parenthesis.

I'm fine with leaving this as is for now for consistency.

qtconsole/console_widget.py
TheMatt2: Updated description, and added a bit more detail

qtconsole/tests/test_00_console_widget.py
TheMatt2: So the idea is kwarg equals doesn't have space? Changed.

TheMatt2:
Merged in main and made format changes. None of the changes in main look like they impact this code directly.

Appreciate you guys taking a look at this.

As for a note on the readiness of this code, I think this code is tested and improves on the current behavior.
The sharpest point that requires some thought is thinking carefully about situations where the "prompt" for the user is a blank string, such as with `input('')`. Since the code uses `prompt_start == prompt_end` to signify there is no prompt at the moment, the code has to be carefully crafted to make sure this isn't an issue. This kind of issue is what is causing #610. For this MR, I made of point of structuring the code in such a way there this should not pose an issue for any situation that the original code could already handle.

qtconsole/console_widget.py
dalthviz: Oh you are totally right! Makes sense to leave it with `\` for consistency :+1:

dalthviz (approved):
Once again thank you @TheMatt2 ! This LGTM :+1:

Just in case, do you have any further feedback @ccordoba12 ?

qtconsole/console_widget.py
```diff
@@ -998,18 +1008,40 @@ def _append_custom(self, insert, input, before_prompt=False, *args, **kwargs):
...
             self._flush_pending_stream()
-            cursor._insert_mode=True
-            cursor.setPosition(self._append_before_prompt_pos)
+
+            # Jump to before prompt, if there is one
+            if cursor.position() >= self._append_before_prompt_pos \
+                    and self._append_before_prompt_pos != self._get_end_pos():
+                cursor.setPosition(self._append_before_prompt_pos)
+
+                # If we appending on the same line as the prompt, use insert mode
```
ccordoba12, line 1020:
```suggestion
                # If we're appending on the same line as the prompt, use insert mode.
```

Fix grammar.

qtconsole/console_widget.py
```diff
@@ -998,18 +1008,40 @@ def _append_custom(self, insert, input, before_prompt=False, *args, **kwargs):
...
+                # If so, the character at self._append_before_prompt_pos will not be a newline
+                cursor.movePosition(QtGui.QTextCursor.Right,
+                                    QtGui.QTextCursor.KeepAnchor)
+                if cursor.selection().toPlainText() != '\n':
+                    cursor._insert_mode = True
+                cursor.movePosition(QtGui.QTextCursor.Left)
         else:
+            # Insert at current printing point
```
ccordoba12, line 1028:
```suggestion
            # Insert at current printing point.
```

Add missing dot.

qtconsole/console_widget.py
```diff
@@ -1657,24 +1689,23 @@ def _event_filter_page_keypress(self, event):
...
-        cursor.movePosition(QtGui.QTextCursor.End)
         self._flush_pending_stream()
-        cursor.movePosition(QtGui.QTextCursor.End)
 
     def _flush_pending_stream(self):
-        """ Flush out pending text into the widget. """
+        """ Flush pending text into the widget. Only applies to text that is pending
+            when the console is in the running state. Text printed when console is
+            not running is shown immediately, and does not wait to be flushed.
+        """
```
ccordoba12, line 1700:
```suggestion
        """
        Flush pending text into the widget.

        It only applies to text that is pending when the console is in the
        running state. Text printed when console is not running is shown
        immediately, and does not wait to be flushed.
        """
```

Improve formatting. I think this won't be done by auto-formatters.

ccordoba12 (commented):
Just some small suggestions for you @TheMatt2, the rest looks good to me.

dalthviz:
@ccordoba12 to move this forward, is okay if we merge this as it is and then over a follow up PR I work on the style suggestions you left here?

ccordoba12:
Sure, go ahead and merge this one.

PR #608, opened 2024-06-04
Use selected syntax style for tracebacks and improve ANSI color codes support

jsbautista:
Fixed traceback higlight, making use of the VerboseTB hack and adding a color homologation to expand the range of colors used by QtConsole
Before:
BeforeQtConsole
BeforeQtConsolea (online-video-cutter com)

After:

AfterQtConsole
AfterQtConsolea

**related issues**
https://github.com/spyder-ide/spyder/issues/4891

### DIFF
--- a/qtconsole/ansi_code_processor.py
+++ b/qtconsole/ansi_code_processor.py
@@ -292,6 +292,31 @@ class AnsiCodeProcessor(object):
             self.actions.append(ScrollAction('scroll', 'down', 'page', 1))
         return ''
 
+    def _parse_ansi_color(self, color, intensity):
+        """
+        Map an ANSI color code to color name or a RGB tuple.
+        Based on: https://gist.github.com/MightyPork/1d9bd3a3fd4eb1a661011560f6921b5b
+        """
+        parsed_color = None
+        if color < 8:
+            # Adjust for intensity, if possible.
+            if intensity > 0:
+                color += 8
+            parsed_color = self.color_map.get(color, None)
+        elif (color > 231):
+                s = int((color - 232) * 10 + 8)
+                parsed_color = (s, s, s)
+        else:
+            n = color - 16
+            b = n % 6
+            g = (n - b) / 6 % 6
+            r = (n - b - g * 6) / 36 % 6
+            r = int(r * 40 + 55) if r else 0
+            g = int(g * 40 + 55) if g else 0
+            b = int(b * 40 + 55) if b else 0
+            parsed_color = (r, g, b)
+        return parsed_color
+
 
 class QtAnsiCodeProcessor(AnsiCodeProcessor):
     """ Translates ANSI escape codes into QTextCharFormats.
@@ -323,12 +348,8 @@ class QtAnsiCodeProcessor(AnsiCodeProcessor):
         """ Returns a QColor for a given color code or rgb list, or None if one
             cannot be constructed.
         """
-
         if isinstance(color, int):
-            # Adjust for intensity, if possible.
-            if color < 8 and intensity > 0:
-                color += 8
-            constructor = self.color_map.get(color, None)
+            constructor = self._parse_ansi_color(color, intensity)
         elif isinstance(color, (tuple, list)):
             constructor = color
         else:
--- a/qtconsole/ansi_code_processor.py
+++ b/qtconsole/ansi_code_processor.py
@@ -300,7 +300,7 @@ class AnsiCodeProcessor(object):
         parsed_color = None
         if color < 16:
             # Adjust for intensity, if possible.
-            if intensity > 0:
+            if intensity > 0 and color < 8:
                 color += 8
             parsed_color = self.color_map.get(color, None)
         elif (color > 231):
--- a/qtconsole/ansi_code_processor.py
+++ b/qtconsole/ansi_code_processor.py
@@ -298,7 +298,7 @@ class AnsiCodeProcessor(object):
         Based on: https://gist.github.com/MightyPork/1d9bd3a3fd4eb1a661011560f6921b5b
         """
         parsed_color = None
-        if color < 8:
+        if color < 16:
             # Adjust for intensity, if possible.
             if intensity > 0:
                 color += 8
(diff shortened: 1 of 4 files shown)


jsbautista:
@dalthviz

dalthviz (commented):
Thanks @jsbautista for the changes here. Left a comment regarding the change over the test. Let me know if you have questions!

qtconsole/mainwindow.py
```diff
@@ -809,7 +810,11 @@ def set_syntax_style(self, syntax_style):
         self.active_frontend._syntax_style_changed()
         self.active_frontend._style_sheet_changed()
         self.active_frontend.reset(clear=True)
-        self.active_frontend._execute("%colors linux", True)
+        self.active_frontend._execute(
+            f"from IPython.core.ultratb import VerboseTB;"
+            "VerboseTB._tb_highlight_style = '{syntax_style}'",
```
ccordoba12, line 815: A couple of comments about this change:

* The `_tb_highlight_style` was added in IPython 8.15 (see https://github.com/ipython/ipython/pull/14138). So, it's necessary a check that executes the new code for that or more recent versions, and leaves the previous one for older versions.
* It'd be a good idea to make `_tb_highlight_style` a public attribute of `VerboseTB` by submitting a PR to IPython. That would require an additional check here but it'd prevent they breaking us in the future.

ccordoba12 (commented):
Thanks @jsbautista and @dalthviz for your work on this!

qtconsole/mainwindow.py
dalthviz: I think something like this could work:

```suggestion
f"""
from IPython.core.ultratb import VerboseTB
try:
    VerboseTB._tb_highlight_style = '{syntax_style}'
except AttributeError:
    get_ipython().run_line_magic('colors', '{colors}')
""",
```

qtconsole/tests/test_jupyter_widget.py
```diff
@@ -34,7 +34,7 @@ def test_stylesheet_changed(self):
         w = JupyterWidget(kind='rich')
 
         # By default, the background is light. White text is rendered as black
-        self.assertEqual(w._ansi_processor.get_color(15).name(), '#000000')
+        self.assertEqual(w._ansi_processor.get_color(15).name(), '#ffffff')
```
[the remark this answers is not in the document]
dalthviz: Note: Missing change revert on the test

qtconsole/mainwindow.py
ccordoba12: This is very good approach, thanks @dalthviz! However, given that `_tb_highlight_style` is a private attribute of `VerboseTB`, I think it'd be a good idea to promote it to a public one in IPython and let Matthias know that we're using it here.

Otherwise, this functionality could break easily without us being aware of it (given that now there's a fallback in case that attribute is no longer available).

dalthviz:
Note: The latest failing tests on linux and mac are unrelated with the work here, seems like a message about the debugger and frozen modules (the one that mentions things like using `PYDEVD_DISABLE_FILE_VALIDATION=1` and `-Xfrozen_modules=off`) is being caught when getting the messages between frontend and kernel

qtconsole/mainwindow.py
dalthviz: Created https://github.com/ipython/ipython/issues/14464 to check your suggestion @ccordoba12

qtconsole/mainwindow.py
```diff
@@ -809,7 +810,16 @@ def set_syntax_style(self, syntax_style):
...
         self.active_frontend._style_sheet_changed()
         self.active_frontend.reset(clear=True)
-        self.active_frontend._execute("%colors linux", True)
+        self.active_frontend._execute(
+f"""
+from IPython.core.ultratb import VerboseTB
+try:
+    VerboseTB._tb_highlight_style = '{syntax_style}'
+except AttributeError:
+    get_ipython().run_line_magic('colors', '{colors}')
```
dalthviz, line 819: Following the response at https://github.com/ipython/ipython/issues/14464#issuecomment-2180056302 I think now this could be:

```suggestion
from IPython.core.ultratb import VerboseTB
if getattr(VerboseTB, 'tb_highlight_style', None) is not None:
    VerboseTB.tb_highlight_style = '{syntax_style}'
elif getattr(VerboseTB, '_tb_highlight_style', None) is not None:
    VerboseTB._tb_highlight_style = '{syntax_style}'
else:
    get_ipython().run_line_magic('colors', '{colors}')
```

ccordoba12: Yep, that should do the trick 👍🏽

qtconsole/mainwindow.py
ccordoba12: Yep, thanks! I saw Matthias answer, so now please proceed to implement his suggestions there.

qtconsole/ansi_code_processor.py
```diff
@@ -292,6 +292,31 @@ def _replace_special(self, match):
...
 
+    def _parse_ansi_color(self, color, intensity):
+        """
+        Map an ANSI color code to color name or a RGB tuple.
+        Based on: https://gist.github.com/MightyPork/1d9bd3a3fd4eb1a661011560f6921b5b
+        """
+        parsed_color = None
+        if color < 16:
+            # Adjust for intensity, if possible.
+            if intensity > 0:
```
dalthviz, line 303: @jsbautista I think we need to also add here a validation to not add `8` to an already brigth code color (`color` >=8; <16). So:
```suggestion
            if intensity > 0 and color < 8:
```

dalthviz (approved):
Thanks @jsbautista for your work here! Although https://github.com/ipython/ipython/pull/14476 is still WIP I'm leaving this approved.

Just in case, do you have any further comments @ccordoba12? Should we wait until https://github.com/ipython/ipython/pull/14476 gets merged to merge this or we could merge this regardless?

ccordoba12:
> Just in case, do you have any further comments @ccordoba12? Should we wait until https://github.com/ipython/ipython/pull/14476 gets merged to merge this or we could merge this regardless?

Nop, I don't have more comments, so please merge it. If https://github.com/ipython/ipython/pull/14476 goes in a different route, we can simply adapt the code here.

PR #611, opened 2024-06-25
PR: Change ANSI color code 3 mapping from `brown` to `gold` and set bold format processing enabled by default

### LINKED ISSUE #273: Unable to output boldface and some colors from executing python script in qtconsole.
How would I go about setting a boldfaced text output? The ansi escape sequence does not work:
print('\033[1m This should be boldface.')
You also cannot use HTML or Markdown as neither of these work in qtconsole.

How would I go about outputting a foreground color of yellow?

print('\033[31m This is red.')
print('\033[32m This is green.')
print('\033[33m This is red but is supposed to be yellow.')

I noticed there is some history regarding ansi escape sequences and making text readable when using light or dark background terminal colors.

Is there a way to set boldface and yellow?

jsbautista:
After:
image
Fixes #273 
Related to https://github.com/spyder-ide/spyder/issues/21225

### DIFF
--- a/qtconsole/tests/test_ansi_code_processor.py
+++ b/qtconsole/tests/test_ansi_code_processor.py
@@ -37,7 +37,7 @@ class TestAnsiCodeProcessor(unittest.TestCase):
     def test_colors(self):
         """ Do basic controls sequences for colors work?
         """
-        string = 'first\x1b[34mblue\x1b[0mlast'
+        string = 'first\x1b[34mblue\x1b[0mlast\033[33mYellow'
         i = -1
         for i, substring in enumerate(self.processor.split_string(string)):
             if i == 0:
@@ -49,6 +49,9 @@ class TestAnsiCodeProcessor(unittest.TestCase):
             elif i == 2:
                 self.assertEqual(substring, 'last')
                 self.assertEqual(self.processor.foreground_color, None)
+            elif i == 3:
+                self.assertEqual(substring, 'Yellow')
+                self.assertEqual(self.processor.foreground_color, 3)
             else:
                 self.fail('Too many substrings.')
         self.assertEqual(i, 2, 'Too few substrings.')
--- a/qtconsole/tests/test_ansi_code_processor.py
+++ b/qtconsole/tests/test_ansi_code_processor.py
@@ -2,13 +2,14 @@
 import unittest
 
 # Local imports
-from qtconsole.ansi_code_processor import AnsiCodeProcessor
+from qtconsole.ansi_code_processor import AnsiCodeProcessor, QtAnsiCodeProcessor
 
 
 class TestAnsiCodeProcessor(unittest.TestCase):
 
     def setUp(self):
         self.processor = AnsiCodeProcessor()
+        self.qt_processor = QtAnsiCodeProcessor()
 
     def test_clear(self):
         """ Do control sequences for clearing the console work?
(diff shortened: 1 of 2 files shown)


jsbautista:
Hi @dalthviz can you review this?

qtconsole/tests/test_ansi_code_processor.py
```diff
@@ -49,6 +49,9 @@ def test_colors(self):
             elif i == 2:
                 self.assertEqual(substring, 'last')
                 self.assertEqual(self.processor.foreground_color, None)
+            elif i == 3:
+                self.assertEqual(substring, 'Yellow')
+                self.assertEqual(self.processor.foreground_color, 3)
             else:
                 self.fail('Too many substrings.')
         self.assertEqual(i, 2, 'Too few substrings.')
```
dalthviz, line 57: Missing an update over the number of substrings here:
```suggestion
        self.assertEqual(i, 3, 'Too few substrings.')
```

dalthviz (commented):
Thanks @jsbautista ! Checked this manually and is working as expected :+1: Just in case, added a suggestion related with the failing test with a possible fix for it.

Also, thinking about tests, probably it could be nice to update the `TestAnsiCodeProcessor` test class here to also have an instance of `QtAnsiCodeProcessor`. In that way we could check the actual color that gets rendered over the console. A possible way to achieve that is:

* Add a `qt_processor` attribute to the `TestAnsiCodeProcessor` of type `QtAnsiCodeProcessor` over the `setUp` method
* Use it over the test you modified to check for the color name with something like:

```python
foreground_color = self.processor.foreground_color
self.assertEqual(self.qt_processor.get_color(foreground_color).name(), '#ffd700')
```

Let me know what you think and if you have any question!

Edit: Also, lets do here the change for the default value to allow bold format and add a test for it. For more details about what needs to be done you can check https://github.com/jupyter/qtconsole/issues/273#issuecomment-2209696367

qtconsole/tests/test_ansi_code_processor.py
```diff
@@ -49,9 +50,15 @@ def test_colors(self):
             elif i == 2:
                 self.assertEqual(substring, 'last')
                 self.assertEqual(self.processor.foreground_color, None)
+            elif i == 3:
+                foreground_color = self.processor.foreground_color
+                self.assertEqual(self.qt_processor.get_color(foreground_color).name(), '#ffd700')
+                self.assertEqual(substring, 'Yellow')
+                self.assertEqual(self.processor.foreground_color, 3)
```
dalthviz, line 57: Just for clarity, maybe could be worthy to write this like:

```suggestion
                foreground_color = self.processor.foreground_color
                self.assertEqual(substring, 'Yellow')
                self.assertEqual(foreground_color, 3)
                self.assertEqual(self.qt_processor.get_color(foreground_color).name(), '#ffd700')
                
```

qtconsole/tests/test_ansi_code_processor.py
```diff
@@ -49,9 +50,16 @@ def test_colors(self):
             elif i == 2:
                 self.assertEqual(substring, 'last')
                 self.assertEqual(self.processor.foreground_color, None)
+            elif i == 3:
+                foreground_color = self.processor.foreground_color
+                self.assertEqual(substring, 'Yellow')
+                self.assertEqual(foreground_color, 3)
+                self.assertEqual(self.qt_processor.get_color(foreground_color).name(), '#ffd700')
+                
+
```
dalthviz, line 59:
```suggestion
```

dalthviz:
Thanks for getting in the suggestions @jsbautista ! Left a single final suggestion to remove some extra blank lines. Also, doing the change to enable bold format by default (following the info at https://github.com/jupyter/qtconsole/issues/273#issuecomment-2209696367) needs to be done. Once those two things are done here I would say this should be ready for merging :+1:

Just in case, what do you think @ccordoba12 ?

dalthviz (approved):
Thanks @jsbautista !

ccordoba12 (approved):
Looks good to me too, thanks @jsbautista!

PR #612, opened 2024-07-19
Remove unnecessary `pyzmq` requirement

xiacunshun:
'pyzmq' is not used in sources, which was pulled in in 79b0b02f3

### DIFF
--- a/requirements/environment.yml
+++ b/requirements/environment.yml
@@ -9,7 +9,6 @@ dependencies:
 - jupyter_client
 - pygments
 - ipykernel
-- pyzmq >=17.1
 
 # For testing
 - coveralls
--- a/setup.py
+++ b/setup.py
@@ -71,7 +71,6 @@ setup_args = dict(
         'pygments',
         'ipykernel>=4.1', # not a real dependency, but require the reference kernel
         'qtpy>=2.4.0',
-        'pyzmq>=17.1',
         'packaging'
     ],
     extras_require = {


xiacunshun:
pyzmq is required by ipykernel and we don't need this in our requirement list.

dalthviz:
Thank you @xiacunshun for your feedback and submitting a PR with the required changes! Checking seems like https://github.com/jupyter/qtconsole/pull/416 was the PR where the `pyzmq` dependency was introduced. To be honest I'm not sure why it was introduced at that time (maybe to make tests past due to some package installation error over CI?). Should the dependency be removed @ccordoba12 ? Seems like tests pass without it. The error over a couple of CI jobs seems like comes from the `Upload coverage to Codecov` step so not related with this 🤔

ccordoba12:
I think this is fine but I don't understand how it'll help your use case @xiacunshun given that IPykernel is (and will continue to be) a dependency of this package.

xiacunshun:
@ccordoba12 I am one of the maintainers of the pipdeptree project and also the maintainer of the OpenCloudOS distribution. Currently, I am developing a Python dependency checking tool. This tool will scan and identify dependencies that do not match the modules actually used in the source code. The aim is to make the dependency relationships of Python projects clearer, which can facilitate us in trimming some unnecessary dependencies when creating images.

xiacunshun:
It's possible that this module may be depended upon by other software, but in reality, it's irrelevant to the current project. If one day the dependent Python project changes its own dependencies, the current project would be unaware of it.

minrk:
pyzmq's a transitive dependency via jupyter-client, etc., so removing it won't remove the dependency, but it may affect the constraint. I think this is looser than others now anyway, so it probably has no effect and can be safely removed. If it's not used directly and if there aren't version _constraints_ specific to this package (there may have been in the past), it doesn't need to be here.

ccordoba12:
Thanks @xiacunshun for the extra info and @minrk for chiming in. I think both arguments make sense so I'm going to merge this PR.

ccordoba12 (approved):
Thanks for your contribution @xiacunshun!

Note: The failures in our tests are unrelated to these changes.

PR #613, opened 2024-07-22
Prevent debugger warning message from appearing while testing communications between kernel and frontend

dalthviz:
Seems like the `test_comms` tests have been failing since when checking communications between kernel and frontend there is a possibility for a debugger warning message to appear (`Debugger warning: It seems that frozen modules are being used...`)


ccordoba12 (approved):
Thanks @dalthviz!

PR #616, opened 2024-07-30
Add logic to handle ANSI escape sequences to move cursor

jsbautista:
I made changes to the way the ANSI code \n is handled. Previously, whenever the \n code appeared, a new line was inserted unconditionally. After the change, a new line is inserted unless there is already a written line at the current position. In that case, instead of inserting a new line, the cursor moves one line down.

Preview with changes made:
image

Comparison of the result obtained with the changes made and in Ipython with QTDM

Preview with changes made:

PreviewTQDMQtConsole

Preview with IPython:

PreviewTQDMIpython

Fixes: #350 
Related with: https://github.com/spyder-ide/spyder-kernels/issues/494 https://github.com/spyder-ide/spyder/issues/6172

### DIFF
--- a/qtconsole/ansi_code_processor.py
+++ b/qtconsole/ansi_code_processor.py
@@ -147,7 +147,7 @@ class AnsiCodeProcessor(object):
 
         if last_char is not None:
             self.actions.append(NewLineAction('newline'))
-            yield last_char
+            yield None
 
     def set_csi_code(self, command, params=[]):
         """ Set attributes based on CSI (Control Sequence Introducer) code.
--- a/qtconsole/ansi_code_processor.py
+++ b/qtconsole/ansi_code_processor.py
@@ -147,7 +147,7 @@ class AnsiCodeProcessor(object):
 
         if last_char is not None:
             self.actions.append(NewLineAction('newline'))
-            yield None
+            yield last_char
 
     def set_csi_code(self, command, params=[]):
         """ Set attributes based on CSI (Control Sequence Introducer) code.
--- a/qtconsole/ansi_code_processor.py
+++ b/qtconsole/ansi_code_processor.py
@@ -92,9 +92,6 @@ class AnsiCodeProcessor(object):
         self.actions = []
         start = 0
 
-        # strings ending with \r are assumed to be ending in \r\n since
-        # \n is appended to output strings automatically.  Accounting
-        # for that, here.
         last_char = None
         string = string[:-1] if last_char is not None else string
 
--- a/qtconsole/ansi_code_processor.py
+++ b/qtconsole/ansi_code_processor.py
@@ -95,7 +95,7 @@ class AnsiCodeProcessor(object):
         # strings ending with \r are assumed to be ending in \r\n since
         # \n is appended to output strings automatically.  Accounting
         # for that, here.
-        last_char = None#'\n' if len(string) > 0 and string[-1] == '\n' else None
+        last_char = None
         string = string[:-1] if last_char is not None else string
 
         for match in ANSI_OR_SPECIAL_PATTERN.finditer(string):
--- a/qtconsole/ansi_code_processor.py
+++ b/qtconsole/ansi_code_processor.py
@@ -122,7 +122,7 @@ class AnsiCodeProcessor(object):
                 self.actions = []
             elif g0 == '\n' or g0 == '\r\n':
                 self.actions.append(NewLineAction('newline'))
-                yield g0
+                yield None
                 self.actions = []
             else:
                 params = [ param for param in groups[1].split(';') if param ]
--- a/qtconsole/ansi_code_processor.py
+++ b/qtconsole/ansi_code_processor.py
@@ -190,10 +190,9 @@ class AnsiCodeProcessor(object):
             count = params[0] if params else 1
             self.actions.append(MoveAction('move', dir, 'line', count))
         elif (command == 'F'):  # Goes back to the begining of the n-th previous line
-            dir = 'up'
+            dir = 'leftup'
             count = params[0] if params else 1
             self.actions.append(MoveAction('move', dir, 'line', count))
-            self.actions.append(CarriageReturnAction('carriage-return'))
         
 
     def set_osc_code(self, params):
--- a/qtconsole/ansi_code_processor.py
+++ b/qtconsole/ansi_code_processor.py
@@ -95,7 +95,7 @@ class AnsiCodeProcessor(object):
         # strings ending with \r are assumed to be ending in \r\n since
         # \n is appended to output strings automatically.  Accounting
         # for that, here.
-        last_char = '\n' if len(string) > 0 and string[-1] == '\n' else None
+        last_char = None#'\n' if len(string) > 0 and string[-1] == '\n' else None
         string = string[:-1] if last_char is not None else string
 
         for match in ANSI_OR_SPECIAL_PATTERN.finditer(string):
--- a/qtconsole/ansi_code_processor.py
+++ b/qtconsole/ansi_code_processor.py
@@ -185,6 +185,17 @@ class AnsiCodeProcessor(object):
             count = params[0] if params else 1
             self.actions.append(ScrollAction('scroll', dir, 'line', count))
 
+        elif (command == 'A'):  # Move N lines Up
+            dir = 'up'
+            count = params[0] if params else 1
+            self.actions.append(MoveAction('move', dir, 'line', count))
+        elif (command == 'F'):  # Goes back to the begining of the n-th previous line
+            dir = 'up'
+            count = params[0] if params else 1
+            self.actions.append(MoveAction('move', dir, 'line', count))
+            self.actions.append(CarriageReturnAction('carriage-return'))
+        
+
     def set_osc_code(self, params):
         """ Set attributes based on OSC (Operating System Command) parameters.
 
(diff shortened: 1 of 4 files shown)


dalthviz:
Thanks for your work here @jsbautista ! As quick comment, adding a test for this new move cursor action logic could be worthy

TheMatt2:
> Preview with changes made in #607: 

I appreciate the goal to add these ANSI codes to allow the cursor to move between  lines. This limitation seems to be the main thing standing between qtconsole being a "true terminal emulator", that is compliant'ish with other terminals.

Though, I see some issues with the changes in this MR.

Your GIF seems to show output that is different that what the intended behavior should be. Here is what your `prueba()` prints in VS Code and MacOS's terminal:

 
image

image

"Segunda prueba" should be printed on the second line, as this is printed before any control characters are occur. Not "Segunda prueba" on the last line.


It also looks like this PR breaks `qtconsole/tests/test_00_console_widget.py::TestConsoleWidget::test_print_carriage_return`

dalthviz:
Hi @TheMatt2 thank you for the feedback and giving this work a check! I think the current GIF being shown is a little bit out of date. The main goal of this PR is to give a handling similar to ANSI codes related with cursor movement as you would observe over IPython (but we could indeed give a check to the behavior of other terminals/platforms to contrast possible differences in behavior). So running the `prueba` code with the latest PR commit on Windows you would get the same output as if you run it with IPython:

* Code:

```python
def prueba():
    print("\rPrimera prueba", end="")
    print("\n", end="")
    print("\rSegunda prueba", end="")
    print("\x1b[A", end="")
    print("\n\n", end="")
    print("\rReemplazo", end="")
    print()
```

* This PR:
image

* IPython:
image

Still, we have some edge cases that need handling and also, as you mention, the changes here are causing some tests as the carriage return one to fail but we are giving all that a check.

Hopefully this makes current state of this PR/work clearer but let us know if you have further suggestions/comments/ideas!

TheMatt2:
>  I think the current GIF being shown is a little bit out of date. The main goal of this PR is to give a handling similar to ANSI codes related with cursor movement as you would observe over IPython (but we could indeed give a check to the behavior of other terminals/platforms to contrast possible differences in behavior). ...

Thank you for your response and clarification. Now I'm with you.

qtconsole/console_widget.py
```diff
@@ -2203,7 +2220,13 @@ def _insert_plain_text(self, cursor, text, flush=False):
...
 
                     elif act.action == 'newline':
-                        cursor.movePosition(QtGui.QTextCursor.EndOfLine)
+                        self.log.debug(self._prompt)
+                        if cursor.block() != cursor.document().lastBlock():
+                            cursor.movePosition(QtGui.QTextCursor.NextBlock)
+                        else:
+                            cursor.movePosition(QtGui.QTextCursor.EndOfLine,
+                                                QtGui.QTextCursor.MoveAnchor)
+                            cursor.insertText('\n')
```
dalthviz, line 2229: @jsbautista I think what is missing here (and the cause of the failed test) is to not only check if we are over the end of the document but also that the document doesn't end with `self._prompt`. So maybe doing something like:

```suggestion
                        if (
                            cursor.block() != cursor.document().lastBlock()
                            and not cursor.document()
                            .toPlainText()
                            .endswith(self._prompt)
                        ):
                            cursor.movePosition(QtGui.QTextCursor.NextBlock)
                        else:
                            cursor.movePosition(
                                QtGui.QTextCursor.EndOfLine,
                                QtGui.QTextCursor.MoveAnchor,
                            )
                            cursor.insertText("\n")
```

Could help tests pass

dalthviz:
Checking aroung some issue related with Rich, using the changes here alone are not enough to see the improvements done. For more details you can check https://github.com/spyder-ide/spyder-kernels/issues/494#issuecomment-2369201113 (where using the Rich Console API is needed - `force_jupyter=False`)

dalthviz (commented):
Thanks @jsbautista for your work here! Left a question and a suggestion. Other than that seems like there is not much else to do here, although maybe could be worthy to do some extra manually testing with things like Rich and TQDM (I already did some while checking https://github.com/spyder-ide/spyder-kernels/issues/494)

qtconsole/ansi_code_processor.py
```diff
@@ -185,6 +182,20 @@ def set_csi_code(self, command, params=[]):
             count = params[0] if params else 1
             self.actions.append(ScrollAction('scroll', dir, 'line', count))
 
+        elif (command == 'A'):  # Move N lines Up
```
ccordoba12, line 185:
```suggestion
        elif command == 'A':  # Move N lines Up
```

No need for using parenthesis here

qtconsole/ansi_code_processor.py
```diff
@@ -185,6 +182,20 @@ def set_csi_code(self, command, params=[]):
...
+            dir = 'up'
+            count = params[0] if params else 1
+            self.actions.append(MoveAction('move', dir, 'line', count))
+        elif (command == 'B'):  # Move N lines Down
```
ccordoba12, line 189:
```suggestion

        elif command == 'B':  # Move N lines Down
```

Follow the format used above these lines.

qtconsole/ansi_code_processor.py
```diff
@@ -185,6 +182,20 @@ def set_csi_code(self, command, params=[]):
...
+            dir = 'down'
+            count = params[0] if params else 1
+            self.actions.append(MoveAction('move', dir, 'line', count))
+        elif (command == 'F'):  # Goes back to the begining of the n-th previous line
```
ccordoba12, line 193:
```suggestion

        elif command == 'F':  # Goes back to the begining of the n-th previous line
```

qtconsole/console_widget.py
```diff
@@ -2188,6 +2188,23 @@ def _insert_plain_text(self, cursor, text, flush=False):
                             cursor.select(QtGui.QTextCursor.Document)
                             cursor.removeSelectedText()
 
+                    elif act.action == 'move' and act.unit == 'line':
+                        if act.dir == 'up':
+                            for i in range(act.count):
+                                cursor.movePosition(
+                                    QtGui.QTextCursor.Up)
```
ccordoba12, line 2195:
```suggestion
                                    QtGui.QTextCursor.Up
                                )
```

Improve code style a bit.

qtconsole/console_widget.py
```diff
@@ -2188,6 +2188,23 @@ def _insert_plain_text(self, cursor, text, flush=False):
...
+                        elif act.dir == 'down':
+                            for i in range(act.count):
+                                cursor.movePosition(
+                                    QtGui.QTextCursor.Down)
```
ccordoba12, line 2199:
```suggestion
                                    QtGui.QTextCursor.Down
                                )
```

qtconsole/console_widget.py
```diff
@@ -2188,6 +2188,23 @@ def _insert_plain_text(self, cursor, text, flush=False):
...
+                        elif act.dir == 'leftup':
+                            for i in range(act.count):
+                                cursor.movePosition(
+                                    QtGui.QTextCursor.Up)
+                            cursor.movePosition(
+                            QtGui.QTextCursor.StartOfLine,
+                            QtGui.QTextCursor.MoveAnchor)
```
ccordoba12, line 2206:
```suggestion
                                    QtGui.QTextCursor.Up
                                )
                            cursor.movePosition(
                                QtGui.QTextCursor.StartOfLine,
                                QtGui.QTextCursor.MoveAnchor
                            )
```

qtconsole/tests/test_ansi_code_processor.py
```diff
@@ -182,6 +182,42 @@ def test_combined(self):
         self.assertEqual(splits, ['abc', None, 'def', None])
         self.assertEqual(actions, [[], ['carriage-return'], [], ['backspace']])
 
+    def test_move_cursor_up(self):
+        string = '\x1b[A\x1b[5A\x1b[F\x1b[5F'
```
ccordoba12, line 186: Please add a comment above this to explain what it does.

ccordoba12 (commented):
Thanks @jsbautista for your work on this! Just some small style suggestions and a comment for you.

qtconsole/tests/test_ansi_code_processor.py
```diff
@@ -186,6 +186,10 @@ def test_move_cursor_up(self):
         """Are the ANSI commands for the cursor movement actions
         (movement up and to the beginning of the line) processed correctly?
         """
+        #This line moves the cursor up once, then moves it up five more lines.
+        #Next, it moves the cursor to the beginning of the previous line, and
+        #finally moves it to the beginning of the fifth line above the current 
+        #position
```
dalthviz, line 192: Lets add a space between the `#` and the text:
```suggestion
        # This line moves the cursor up once, then moves it up five more lines.
        # Next, it moves the cursor to the beginning of the previous line, and
        # finally moves it to the beginning of the fifth line above the current 
        # position
```

dalthviz (approved):
Thanks @jsbautista for all your work here! This LGTM :+1:

Just in case, do you have any further comments @ccordoba12 ? Also, in case this is ready to be merged, should the millestone be set to be `5.7.0` maybe (since this adds new functionality)?

ccordoba12 (approved):
Looks good to me now, thanks @jsbautista and @dalthviz for this great improvement!

ccordoba12:
> Also, in case this is ready to be merged, should the millestone be set to be 5.7.0 maybe (since this adds new functionality)?

I think this is closer to be a bug fix than new functionality because it doesn't add new config options, features or shortcuts. As you can see in our [changelog](https://github.com/jupyter/qtconsole/blob/main/docs/source/changelog.rst), we release a new minor version when there are additions like those to the package, or when dropping Python versions.

ccordoba12:
Also, since this basically fixes https://github.com/spyder-ide/spyder/issues/6172, a PR in Spyder should be created to sync our Qtconsole subrepo there with these changes.

PR #618, opened 2024-08-12
Missing code style suggestions from Carriage Return fix PR

dalthviz:
Follow up for #607 as discussed at https://github.com/jupyter/qtconsole/pull/607#issuecomment-2276530999

### DIFF
--- a/qtconsole/console_widget.py
+++ b/qtconsole/console_widget.py
@@ -1017,7 +1017,7 @@ class ConsoleWidget(MetaQObjectHasTraits('NewBase', (LoggingConfigurable, superQ
                     and self._append_before_prompt_pos != self._get_end_pos():
                 cursor.setPosition(self._append_before_prompt_pos)
 
-                # If we're appending on the same line as the prompt, use insert mode
+                # If we're appending on the same line as the prompt, use insert mode.
                 # If so, the character at self._append_before_prompt_pos will not be a newline
                 cursor.movePosition(QtGui.QTextCursor.Right,
                                     QtGui.QTextCursor.KeepAnchor)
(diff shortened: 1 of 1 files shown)


qtconsole/console_widget.py
```diff
@@ -1694,9 +1694,12 @@ def _on_flush_pending_stream_timer(self):
         self._flush_pending_stream()
 
     def _flush_pending_stream(self):
-        """ Flush pending text into the widget. Only applies to text that is pending
-            when the console is in the running state. Text printed when console is
-            not running is shown immediately, and does not wait to be flushed.
+        """
+        Flush pending text into the widget.
+
+        Only applies to text that is pending when the console is in the
```
ccordoba12, line 1700:
```suggestion
        It only applies to text that is pending when the console is in the
```

qtconsole/console_widget.py
```diff
@@ -1017,15 +1017,15 @@ def _append_custom(self, insert, input, before_prompt=False, *args, **kwargs):
                     and self._append_before_prompt_pos != self._get_end_pos():
                 cursor.setPosition(self._append_before_prompt_pos)
 
-                # If we appending on the same line as the prompt, use insert mode
+                # If we're appending on the same line as the prompt, use insert mode
```
ccordoba12, line 1020:
```suggestion
                # If we're appending on the same line as the prompt, use insert mode.
```

ccordoba12 (commented):
Two tiny suggestions that I missed in my previous review, the rest looks good to me.

ccordoba12 (approved):
Thanks @dalthviz!

TheMatt2:
I am total on board with these changes. Thanks for taking care of this. I've been busy with other things.

PR #621, opened 2024-09-02
PR: Changes for handling/customizing shortcuts via traitlet variables and a coverage upload fix

jsbautista:
* Adds a way to customize actions shortcuts via traitlet variables
* Fixes coverage upload flakiness

Preview:
PreviewShortCuts


qtconsole/console_widget.py
```diff
@@ -43,6 +45,17 @@ def is_whitespace(char):
...
+        # Define traits for shortcuts
+        shortcut_print = Unicode('Ctrl+P').tag(config=True)
+        shortcut_select_all = Unicode('Ctrl+A').tag(config=True)
+        shortcut_cut = Unicode(QtGui.QKeySequence(QtGui.QKeySequence.Cut).toString()).tag(config=True)
+        shortcut_copy = Unicode(QtGui.QKeySequence(QtGui.QKeySequence.Copy).toString()).tag(config=True)
+        shortcut_paste = Unicode(QtGui.QKeySequence(QtGui.QKeySequence.Paste).toString()).tag(config=True)
+
+        @observe('shortcut_print', 'shortcut_select_all','shortcut_cut','shortcut_copy','shortcut_paste')
+        def _on_shortcut_changed(self, change):
+            print(f"Shortcut for {change['name']} changed to: {change['new']}")
```
dalthviz, line 58: So checking locally, I think the reason why adding this class causes a segfault is due to the usage of `QKeySequence` to define default values. The documentation of `QKeySequence` mentions that [if there is not a `QApplication` instance initialized calling `QKeySequence` can cause your application to crash](https://doc.qt.io/qt-6/qkeysequence.html#gnu-emacs-style-key-sequences). To prevent this I think we could use the traitlet `default` decorator. So this class definition will end up looking something like:

```suggestion

class ShortcutManager(HasTraits):
    """Default shortcuts definition and changes event handler."""

    # Define traits for shortcuts
    shortcut_print = Unicode('Ctrl+P').tag(config=True)
    shortcut_select_all = Unicode('Ctrl+A').tag(config=True)
    shortcut_cut = Unicode().tag(config=True)
    shortcut_copy = Unicode().tag(config=True)
    shortcut_paste = Unicode().tag(config=True)

    @default('shortcut_cut')
    def _default_shortcut_cut(self):
        return QtGui.QKeySequence(QtGui.QKeySequence.Cut).toString()

    @default('shortcut_copy')
    def _default_shortcut_copy(self):
        return QtGui.QKeySequence(QtGui.QKeySequence.Copy).toString()

    @default('shortcut_paste')
    def _default_shortcut_paste(self):
        return QtGui.QKeySequence(QtGui.QKeySequence.Paste).toString()

    @observe('shortcut_print', 'shortcut_select_all','shortcut_cut','shortcut_copy','shortcut_paste')
    def _on_shortcut_changed(self, change):
        print(f"Shortcut for {change['name']} changed to: {change['new']}")


```

qtconsole/console_widget.py
```diff
@@ -21,6 +21,8 @@
 
 from traitlets.config.configurable import LoggingConfigurable
 from traitlets import Bool, Enum, Integer, Unicode
+from traitlets import HasTraits, Unicode, observe
+
```
dalthviz, line 25: Need to import `default` for the suggestion about the `ShortcutManager` class definition works:
```suggestion
from traitlets import default, HasTraits, Unicode, observe
```

qtconsole/console_widget.py
```diff
@@ -319,19 +333,22 @@ def __init__(self, parent=None, **kw):
         # Set a monospaced font.
         self.reset_font()
 
+        self.shortcut_manager = ShortcutManager()
+
         # Configure actions.
         action = QtWidgets.QAction('Print', None)
         action.setEnabled(True)
-        printkey = QtGui.QKeySequence(QtGui.QKeySequence.Print)
+        printkey = self.shortcut_manager.shortcut_print
```
dalthviz, line 341: After applying locally the suggestions above an error appeared regarding the validation done below (using `QKeySequence.matches`). Due to that, here we need to keep `printkey` as a `QKeySequence` instance so this should look something like:
```suggestion
        printkey = QtGui.QKeySequence(self.shortcut_manager.shortcut_print)
```

qtconsole/console_widget.py
```diff
@@ -342,16 +359,17 @@ def __init__(self, parent=None, **kw):
 
         action = QtWidgets.QAction('Select All', None)
         action.setEnabled(True)
-        selectall = QtGui.QKeySequence(QtGui.QKeySequence.SelectAll)
+        selectall = self.shortcut_manager.shortcut_select_all
```
dalthviz, line 362: Similar to the comment above, `selectall` needs to be still a `QKeySequence` instance to be able to use `QKeySequence.matches` in the validation below this line. So this should look something like:
```suggestion
        selectall = QtGui.QKeySequence(self.shortcut_manager.shortcut_select_all)
```

dalthviz (commented):
Thank @jsbautista for your work here! Added some suggestions that should prevent the changes here to cause a segfault. If you have any question about the suggested changes let me know!

dalthviz (commented):
Thanks for the latest changes @jsbautista ! Left some new comments/suggestions related with the method that gathers the different actions with shortcuts over a dictionary. Also, it could be nice to add a GIF or a command/flag/config file over the PR description to show an example of how changing the shortcuts could be done using the work being done here.

dalthviz:
Note: Adding this as part of the 5.7.0 release but depending on reviews/feedback probably could be moved to another future minor/major release (5.8.0 or 6.0.0)

PR #622, opened 2024-10-01
PR: Create widget to manage list history

jsbautista:
Preview:
PreviewDropListHistory


dalthviz:
Note: Seems like Linux tests jobs are failing due to missing windows manager config (which then makes a check for widget visibility fail)? 🤔

dalthviz:
Note: Merging this should be done via a squash merge (there are quite a number of commits with duplicated/not very informative messages)

dalthviz:
Do you have any comments/feedback on this work @ccordoba12 ? I think it looks good from my side but before merging it I think it could be nice to have this work reviewed by someone else 🤔

dalthviz:
Note: Adding this as part of the 5.7.0 release but depending on reviews/feedback probably could be moved to another future minor/major release (5.8.0 or 6.0.0)

PR #627, opened 2025-01-03
Update setup-miniconda action version and use miniforge due to mambaforge sunsetting

dalthviz:
See https://github.com/jupyter/qtconsole/actions/runs/12600493770/job/35119642403?pr=622#step:4:128 and https://conda-forge.org/news/2024/07/29/sunsetting-mambaforge/


ccordoba12 (approved):
Thanks @dalthviz!

PR #628, opened 2025-01-16
IPython pygments lexers are moving to separate package

takluyver:
See https://github.com/ipython/ipython/issues/14520. The newly split package is already available now; I plan to leave the `IPython.lib.lexers` import working for a while, but I thought I'd update this while I'm thinking about it.

### DIFF
--- a/.github/workflows/linux-tests.yml
+++ b/.github/workflows/linux-tests.yml
@@ -33,6 +33,8 @@ jobs:
         exclude:
           - INSTALL_TYPE: 'conda'
             QT_LIB: 'pyqt6'
+          - INSTALL_TYPE: 'conda'
+            PYTHON_VERSION: '3.8'
     timeout-minutes: 15
     steps:
       - name: Checkout branch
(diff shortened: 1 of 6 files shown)


takluyver:
I'll probably extend the Python compatibility of the new package back some way. See also https://github.com/spyder-ide/spyder/pull/23481

takluyver:
The Python 3.8 tests using conda are still failing. I've used conda-forge's `python_min` variable in ipython_pygments_kernels, so it's only available in conda-forge for Python >=3.9.

Should I look into if there's a way round that, or is it time for qtconsole to drop Python 3.8? Or to drop testing with conda on 3.8?

takluyver:
Thanks! I've removed the conda builds using Python 3.8 as you suggested, leaving the Linux + pip test jobs (pyqt5 & pyqt6) as the only remaining ones on 3.8.

Carreau:
Merging as @ccordoba12 seem to be busy.

PR #629, opened 2025-02-12
Test  async kernel manager

Carreau:
In ipykernel 7.0 start_kernel can be an async function.

### DIFF
--- a/.github/workflows/linux-tests.yml
+++ b/.github/workflows/linux-tests.yml
@@ -27,7 +27,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        PYTHON_VERSION: ['3.8', '3.9', '3.10', '3.11']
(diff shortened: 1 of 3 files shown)


ccordoba12:
> This PR add a test for this, and make qtconsole compatible.

I don't understand the compatibility part because this PR only adds a test for the async kernel manager.

Carreau:
> I don't understand the compatibility part because this PR only adds a test for the async kernel manager.

Sorry the part about compatibility was removed, I'll do it separately. 

ipykernel will change at some point in the future so that method on the base class will be async, so it might need extra changes here, but it's a bit complicated to to in one pass.

ccordoba12 (approved):
Thanks @Carreau!

Carreau:
Thanks !

PR #633, opened 2025-06-19
PR: Fix async kernel test when using conda (CI)

dalthviz:
Fixes issue described mentioned over https://github.com/spyder-ide/qtconsole/pull/632#issuecomment-2986518442


ccordoba12 (approved):
Thanks @dalthviz!

PR #639, opened 2025-07-29
Check for kernel client when showing interpreter

mrclary:
Check for kernel client before setting `_prompt_requested` or sending an execute command in `JupyterWidget._show_interpreter_prompt`.

Fixes an issue where the prompt may not be displayed.
See spyder-ide/spyder#24269


jitseniesen (approved):

ccordoba12 (approved):
Thanks @mrclary!

PR #646, opened 2026-01-15
Single-source CI jobs & use alls-green to check status

CAM-Gerlach:
Note: This PR is from an upstream branch since newly-added GH Actions workflows won't kick off from a fork branch; see PR #645 for the previous failed attempt.

We currently have three GH Actions workflow files to run the test, one per platform, each with largely duplicative content which is not very DRY. Additionally, the number of matrix jobs make it a pain to keep on top of setting the correct jobs as required given the serious limitations of GitHub's interface in that regard.

Therefore, this PR single-sources the various jobs to all run via a single workflow without duplication, and uses the [alls-green action](https://github.com/re-actors/alls-green) to check their required status, like we do for CPython, so that only that job need be required and it will ensure the matrix jobs are required automatically.

It also refactors the workflow to fix some other minor issues and cruft:
* Set the Windows and Mac upper bound version to the actual supported upper bound (3.13, not 3.12)
* Switch the Coveralls action to run on a supported Python version so it actually works (it previously was set to run on 3.8 and never updated when that was removed from the matrix
* Remove an old Python 3.8 matrix exclude that is no longer relevant for the same reason
* Use the job matrix to centrally manage config for the coveralls job and the pyqt6 install command instead of scattered around different steps
* Add `conda-remove-defaults: true` to address an outstanding warning
* Update the checkout action from v3 to the latest v6
* Set the default run invocation to `bash -el {0}` instead of duplicating it every step
* Other minor cleanup

### DIFF
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -45,10 +45,6 @@ jobs:
               REQUIREMENTS_FILE: 'requirements/environment.yml'
             - INSTALL_TYPE: 'pip'
               REQUIREMENTS_FILE: 'requirements/pip.yml'
-            - QT_LIB: 'pyqt5'
-              QT_DEPS: 'pyqt5'
-            - QT_LIB: 'pyqt6'
-              QT_DEPS: 'pyqt6!=6.4.0 pyqt6-qt6!=6.4.0'
             - RUNNER_OS: 'ubuntu'
               PYTHON_VERSION: '3.13'
               COVERAGE: 'True'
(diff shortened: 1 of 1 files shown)


CAM-Gerlach:
@ccordoba12 I recommend merging it as "Merge" rather than "Squash", since I've structured things into atomic commits and it makes it much easier to see what has actually been changed and added versus the previous CI workflows rather than one big commit deleting the old and adding a brand new one.

ccordoba12 (approved):
Great work, thanks @CAM-Gerlach!

PR #650, opened 2026-02-10
Handle ANSI codes to show/hide cursor

### LINKED ISSUE #648: Handling of ansi code ?25h
Currently the rich progress bar is not working entirely in the spyder console. A minimal example:
```
import sys
import time
print(f'{sys.stdout.isatty()=}') # True in Spyder
from rich.progress import track

for j in track([1,2,3,4]):
    time.sleep(.2);
    print(j)
    
print('--')    

# works, expect for printing the odd character [?25h (could be "make cursor visible", see https://gist.github.com/fnky/458719343aabd01cfb17a3a4f7296797)
```
outputs the progress bar, but with a strange character at the end. The character seems to be unicode `ESC[?25h` (e.g. `make cursor visible`).
This snippet gives the same result in the Spyder console:
```
f = "\x1b[?25h"
print(f'-{f}-')
```
Can we add support for this ansi code to spyder?

jsbautista:
Fixes #648.

### DIFF
--- a/qtconsole/ansi_code_processor.py
+++ b/qtconsole/ansi_code_processor.py
@@ -144,6 +144,7 @@ class AnsiCodeProcessor(object):
 
                 elif g0.startswith(']'):
                     # Case 2: OSC code.
+                    params = [ param for param in groups[1].split(';') if param ]
                     self.set_osc_code(params)
 
         raw = string[start:]
--- a/qtconsole/ansi_code_processor.py
+++ b/qtconsole/ansi_code_processor.py
@@ -125,9 +125,14 @@ class AnsiCodeProcessor(object):
                 yield None
                 self.actions = []
             else:
-                params = [ param for param in groups[1].split(';') if param ]
                 if g0.startswith('['):
-                    # Case 1: CSI code.
+                    raw_params = groups[1] or ""
+                    # Handle private mode sequences
+                    if raw_params.startswith('?'):
+                        raw_params = raw_params[1:]
+
+                    params = [p for p in raw_params.split(';') if p]
+
                     try:
                         params = list(map(int, params))
                     except ValueError:
--- a/qtconsole/ansi_code_processor.py
+++ b/qtconsole/ansi_code_processor.py
@@ -125,6 +125,7 @@ class AnsiCodeProcessor(object):
                 yield None
                 self.actions = []
             else:
+                params = []
                 if g0.startswith('['):
                     raw_params = groups[1] or ""
                     # Handle private mode sequences
--- a/qtconsole/ansi_code_processor.py
+++ b/qtconsole/ansi_code_processor.py
@@ -18,6 +18,9 @@ from qtconsole.styles import dark_style
 # Constants and datatypes
 #-----------------------------------------------------------------------------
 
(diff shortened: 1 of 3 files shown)


wmvanvliet:
How easy would it now be for the terminal to remain in a state where the cursor is permanently hidden? For example, when the output of a command is cut off, or is malformed in some way. It would be good to also have a way to reset the state back to the default. Perhaps when we issue a "clear screen" (`CTRL+L`) command?

ccordoba12:
> It would be good to also have a way to reset the state back to the default. Perhaps when we issue a "clear screen" (CTRL+L) command?

Thanks for chiming in @wmvanvliet! I think that's a good idea.

qtconsole/tests/test_ansi_code_processor.py
```diff
@@ -224,7 +224,28 @@ def test_move_cursor_up(self):
...
+        """Test for the ANSI commands for cursor visibility (?25h / ?25l)
+        """
+        # This line hides the cursor, then shows it again.
+        string = '\x1b[?25l\x1b[?25h'
+        i = -1
+
+        for i, substring in enumerate(self.processor.split_string(string)):
+            self.assertEqual(len(self.processor.actions), 1)
+            action = self.processor.actions[0]
+
```
ccordoba12:
```suggestion

            action = self.processor.actions[0]
```

Improve code style a bit.

qtconsole/ansi_code_processor.py
```diff
@@ -18,6 +18,9 @@
 # Constants and datatypes
 #-----------------------------------------------------------------------------
 
+# An action for cursor visibility requests
+CursorVisibilityAction = namedtuple('CursorVisibilityAction', ['action', 'visible'])
```
ccordoba12:
```suggestion
CursorVisibilityAction = namedtuple(
    'CursorVisibilityAction', ['action', 'visible']
)
```

Fix code style.

qtconsole/ansi_code_processor.py
```diff
@@ -122,9 +125,15 @@ def split_string(self, string):
                 yield None
                 self.actions = []
             else:
-                params = [ param for param in groups[1].split(';') if param ]
+                params = []
                 if g0.startswith('['):
-                    # Case 1: CSI code.
+                    raw_params = groups[1] or ""
+                    # Handle private mode sequences
```
ccordoba12, line 134:
```suggestion

                    # Handle private mode sequences
```

For clarity.

qtconsole/ansi_code_processor.py
```diff
@@ -135,6 +144,7 @@ def split_string(self, string):
 
                 elif g0.startswith(']'):
                     # Case 2: OSC code.
+                    params = [ param for param in groups[1].split(';') if param ]
```
ccordoba12:
```suggestion
                    params = [param for param in groups[1].split(';') if param]
```

Remove unnecessary blanks.

ccordoba12 (commented):
Just a few suggestions for you @jsbautista.

ccordoba12 (approved):
Looks good to me now, thanks @jsbautista!

